PR #111, opened 2021-08-16
test: Run toml-test

epage:
We are skipping all tests that `toml_edit` fails with the expectation
that we'll use the ignores as the task list for what is left towards
toml 1.0 and re-enable them as they start passing.

See https://github.com/BurntSushi/toml-test for details

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,11 @@ combine = "4.5.2"
 [dev-dependencies]
 serde_json = "1.0.44"
 pretty_assertions = "0.7.1"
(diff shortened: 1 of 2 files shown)


epage:
I've have several branches that start to whittle down the list of ignores, starting with the easy ones to help bring me up to speed on the code base and combine (I'm more of a nom person).

I can add all of the commits to this PR, put up a PR with all of the commits after this, or put up a PR one at a time for each set of tests that starts passing.  Let me know what you're preferred approach is.

PR #115, opened 2021-08-23
fix!: Allow mixed-type arrays

epage:
BREAKING CHANGE: Array functions that returned a `Result` to report
mixed-type arrays no longer do so.


ordian (approved):
Huh, what once was a feature now is a bug

PR #118, opened 2021-08-23
test: Enable underscore test

epage:
The root cause of this was a bug in the test harness.


epage:
This was so minor, I just went ahead and merged it.  Hope that is ok

PR #215, opened 2021-09-24
perf: Remove vec alloc on every key-value pair

epage:
We represent all paths as `Vec1<Key>`, requiring allocating at least one
element for every key-value pair.  This change gets us a stack-value for
the one element, so in the majority of cases, we bypass the heap.


epage:
This had no impact on performance.

PR #276, opened 2021-12-28
Add `Table::sort_by` to support custom sorting

bryanburgers:
Also add `InlineTable::sort_by` for parity.

We have large TOML documents that have been sorted using a
non-TOML-aware sorter (for example, highlighting a block of code and
using VSCode to sort a table). This ends up sorting quoted keys ahead of
non-quoted keys.

```toml
[table]
"a key" = 1
"b key" = "Example"
"c key" = false
akey = 2
bkey = "Another example"
ckey = true
```

When adding a new entry to a table, we would like to keep that sort. The
current `Table::sort_values` function intelligently sorts by the value
of the key.

```toml
[table]
"a key" = 1
akey = 2
"b key" = "Example"
bkey = "Another example"
"c key" = false
ckey = true
```

To allow controlling the sort order, add `Table::sort_by` (which is a
thin wrapper over `IndexMap::sort_by`). This allows us to do the
following to preserve non-intelligent sorting.

```rust
table.insert("another", Item::Value("Another".into()));
table.sort_by(|k1, _, k2, _| k1.to_repr().as_raw().cmp(k2.to_repr().as_raw());
```

And get the output that we want.

```toml
[table]
"a key" = 1
"b key" = "Example"
"c key" = false
akey = 2
another = "Another"
bkey = "Another example"
ckey = true
```

### DIFF
--- a/src/inline_table.rs
+++ b/src/inline_table.rs
@@ -97,32 +97,26 @@ impl InlineTable {
     ///
     /// The comparison function receives two key and value pairs to compare (you can sort by keys or
     /// values or their combination as needed).
-    pub fn sort_values_by<F>(&mut self, compare: F)
+    pub fn sort_values_by<F>(&mut self, mut compare: F)
     where
-        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
+        F: FnMut(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
     {
-        let cmp = &compare;
+        self.sort_values_by_internal(&mut compare);
+    }
+
+    fn sort_values_by_internal<F>(&mut self, compare: &mut F)
+    where
+        F: FnMut(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
+    {
         let modified_cmp = |_: &InternalString,
                             val1: &TableKeyValue,
                             _: &InternalString,
                             val2: &TableKeyValue|
          -> std::cmp::Ordering {
-            cmp(&val1.key, &val1.value, &val2.key, &val2.value)
+            compare(&val1.key, &val1.value, &val2.key, &val2.value)
         };
 
-        self.sort_values_by_internal(&modified_cmp);
-    }
-
-    fn sort_values_by_internal<F>(&mut self, compare: &F)
-    where
-        F: Fn(
-            &InternalString,
-            &TableKeyValue,
-            &InternalString,
-            &TableKeyValue,
-        ) -> std::cmp::Ordering,
-    {
-        self.items.sort_by(compare);
+        self.items.sort_by(modified_cmp);
         for kv in self.items.values_mut() {
             match &mut kv.value {
                 Item::Value(Value::InlineTable(table)) if table.is_dotted() => {
--- a/src/inline_table.rs
+++ b/src/inline_table.rs
@@ -93,6 +93,46 @@ impl InlineTable {
         }
     }
 
+    /// Sort Key/Value Pairs of the table using the using the comparison function `compare`.
+    ///
+    /// The comparison function receives two key and value pairs to compare (you can sort by keys or
+    /// values or their combination as needed).
+    pub fn sort_by<F>(&mut self, compare: F)
+    where
+        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
+    {
+        let cmp = &compare;
+        let modified_cmp = |_: &InternalString,
+                            val1: &TableKeyValue,
+                            _: &InternalString,
+                            val2: &TableKeyValue|
+         -> std::cmp::Ordering {
+            cmp(&val1.key, &val1.value, &val2.key, &val2.value)
+        };
+
+        self.sort_by_internal(&modified_cmp);
+    }
+
+    fn sort_by_internal<F>(&mut self, compare: &F)
+    where
+        F: Fn(
+            &InternalString,
+            &TableKeyValue,
+            &InternalString,
+            &TableKeyValue,
+        ) -> std::cmp::Ordering,
+    {
+        self.items.sort_by(compare);
+        for kv in self.items.values_mut() {
+            match &mut kv.value {
+                Item::Value(Value::InlineTable(table)) if table.is_dotted() => {
+                    table.sort_by_internal(compare);
+                }
+                _ => {}
+            }
+        }
+    }
+
     /// Change this table's dotted status
     pub fn set_dotted(&mut self, yes: bool) {
         self.dotted = yes;
--- a/src/inline_table.rs
+++ b/src/inline_table.rs
@@ -97,7 +97,7 @@ impl InlineTable {
     ///
     /// The comparison function receives two key and value pairs to compare (you can sort by keys or
     /// values or their combination as needed).
-    pub fn sort_by<F>(&mut self, compare: F)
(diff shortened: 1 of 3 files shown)


src/inline_table.rs
```diff
@@ -93,6 +93,46 @@ impl InlineTable {
         }
     }
 
+    /// Sort Key/Value Pairs of the table using the using the comparison function `compare`.
+    ///
+    /// The comparison function receives two key and value pairs to compare (you can sort by keys or
+    /// values or their combination as needed).
+    pub fn sort_by<F>(&mut self, compare: F)
```
epage, line 100: Could you rename the functions to `sort_values_by` for consistency?

src/inline_table.rs
```diff
@@ -93,6 +93,46 @@ impl InlineTable {
...
+    where
+        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
```
epage, line 102: `Item` or `Value`?

src/table.rs
```diff
@@ -117,6 +117,46 @@ impl Table {
         }
     }
 
+    /// Sort Key/Value Pairs of the table using the using the comparison function `compare`.
+    ///
+    /// The comparison function receives two key and value pairs to compare (you can sort by keys or
+    /// values or their combination as needed).
+    pub fn sort_by<F>(&mut self, compare: F)
+    where
+        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
```
epage, line 126: We should probably expose this in `TableLike` as well

epage (commented):
Thanks for adding this!

Personally, I'm a little torn on how much we should copy `IndexMap` for this but figure we can release this and see how ti works out.

src/inline_table.rs
bryanburgers: Done!

src/table.rs
bryanburgers: Oh boo. Doing this would cause object safety issues. Here's what the compiler tells me when I try to do that.

```
   Compiling toml_edit v0.12.0 (/Users/bryan/third-party/toml_edit)
error[E0038]: the trait `TableLike` cannot be made into an object
   --> src/item.rs:254:36
    |
254 |     pub fn as_table_like(&self) -> Option<&dyn TableLike> {
    |                                    ^^^^^^^^^^^^^^^^^^^^^^ `TableLike` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> src/table.rs:486:8
    |
447 | pub trait TableLike: crate::private::Sealed {
    |           --------- this trait cannot be made into an object...
...
486 |     fn sort_values_by<F>(&mut self, compare: F)
    |        ^^^^^^^^^^^^^^ ...because method `sort_values_by` has generic type parameters
    = help: consider moving `sort_values_by` to another trait
```

src/inline_table.rs
bryanburgers: We're ultimately sorting on a `TableKeyValue`, which has a `Key` and an `Item`. That's why I went with `Item`.

Is there a guarantee that the `Item` will necessarily always be a `Value`? If so, we can sort on `val.value.as_value().unwrap()` instead of `&val.value`.

Your call.

src/table.rs
```diff
@@ -117,6 +117,46 @@ impl Table {
         }
     }
 
+    /// Sort Key/Value Pairs of the table using the using the comparison function `compare`.
+    ///
+    /// The comparison function receives two key and value pairs to compare (you can sort by keys or
+    /// values or their combination as needed).
+    pub fn sort_values_by<F>(&mut self, compare: F)
+    where
+        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
```
epage, line 126: There a reason you do `Fn` unlike `IndexMap::solrt_by`s `FnMut`?

src/table.rs
epage: Good point.  Not saying we have to do this but one option is
```rust
pub type FnCmp<'a> = dyn Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering + 'a;
pub(crate) type FnCmpInternal<'a> = dyn Fn(&InternalString, &TableKeyValue, &InternalString, &TableKeyValue) -> std::cmp::Ordering
    + 'a;

...

    pub fn sort_values_by<'a>(&mut self, compare: &FnCmp<'a>) {
    }

    fn sort_values_by_internal<'a>(&mut self, compare: &FnCmpInternal<'a>) {
    }

```

src/inline_table.rs
epage: Only through developers abusing corners in our API do we allow an `InlineTable` to have an item.  We then pretty much ignore any non-`Value` `Item` in the table (we won't render it).

I wouldn't go as far as `.unwrap()` but I would choose an arbitrary sort policy for it.

src/table.rs
bryanburgers: Yeah. The reason is that I tried a bunch of things, but I couldn't get anything with `FnMut` to type check properly.

bryanburgers: OK, I took another run at it and I was able to get it this time. 👍

src/inline_table.rs
```diff
@@ -97,32 +97,26 @@ impl InlineTable {
...
+    pub fn sort_values_by<F>(&mut self, mut compare: F)
     where
-        F: Fn(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
+        F: FnMut(&Key, &Item, &Key, &Item) -> std::cmp::Ordering,
+    {
+        self.sort_values_by_internal(&mut compare);
+    }
+
+    fn sort_values_by_internal<F>(&mut self, compare: &mut F)
+    where
```
epage, line 108: Do we still need `_internal`?

bryanburgers: Yes. Because we want to recursively sort in a loop, we either need to

1. add `where F: Copy` as a restriction on the function, or
2. take a `&mut F` reference instead of taking ownership.

`F: Copy` seems too restrictive (like you mentioned previously, we want to get as close to `IndexMap::sort_by`'s requirements as possible: just an `FnMut`)

Changing the signature to `Table::sort_values_by<F>(&mut self, compare: &mut F)` is an option – and then we don't need `_internal`, but then the signature differs from the expectations that most people have from things like `Vec::sort_by` and `IndexMap::sort_by`.

And even though `&mut F` implements `FnMut`, we can't take ownership in `Table::sort_values_by(&mut self, mut compare: F)` and then make the recursive call with `table.sort_values_by(&mut compare)` because then Rust detects an infinite loop trying to do type inference.

So basically, the way to get it done is for the signature to be `Table::some_function(&mut self, compare: &mut F)` and the recursion to be `sub_table.some_function(compare /* pass along the &mut F */)`. `Table::sort_values_by_internal` fills that roll, and `Table::sort_values_by` provides a better interface to it.

epage: Ah, missed the subtle signature difference, thanks!

epage:
I'm going to go ahead and merge.  I'll play with it to see if I can get the other parts working, otherwise I'll release it as-is.

epage:
v0.12.1 is released!

bryanburgers:
Awesome! Thank you!

PR #290, opened 2022-01-07
docs: Remove reference to removed function

### LINKED ISSUE #289: Docs reference to_string_in_original_order
In https://github.com/ordian/toml_edit/blob/a06c6f5a294155d39cdd06a586c6969027bdaea1/src/table.rs#L191-L197, the docs reference `Document::to_string_in_original_order` which seems to be gone.

I'm don't know how to best rewrite the text as I'm not sure if there are even two cases left or `set_position` is always respeced.

epage:
It is now the only behavior

Fixes #289

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -189,9 +189,6 @@ impl Table {
     }
 
     /// Sets the position of the `Table` within the `Document`.
-    ///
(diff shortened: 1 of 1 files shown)


webmaster128 (approved):
Thanks!

Issue #291, opened 2022-01-07
Insert at specific position


webmaster128:
The `Table` positions are documented to only be set when they are parsed from a document. Is it possible in insert a table at a specific place of a document?

E.g. if I want to add a `[profile.custom]` before or after `[profile.release]` in https://github.com/CosmWasm/cosmwasm/blob/v1.0.0-beta4/contracts/hackatom/Cargo.toml#L13-L22, is that possible?

Right now I do

```rs
                let mut profile = toml_edit::Table::new();
                profile.set_implicit(true);
                profile.insert("custom", toml_edit::Item::Table(toml_edit::Table::new()));
                doc.insert("profile", toml_edit::Item::Table(profile));
```

which appends `[profile.custom]` at the end of the document.

epage:
> The Table positions are documented to only be set when they are parsed from a document. Is it possible in insert a table at a specific place of a document?

Which piece of documentation are you referring to?  `set_position` exists to be able to control the ordering of **all** tables within a document.

webmaster128:
Ah, cool. I guess it was a combination of

```rs
    // used for putting tables back in their original order when serialising.
    // Will be None when the Table wasn't parsed from a file.
    pub(crate) position: Option<usize>,
```

```rs
    /// Sets the position of the `Table` within the `Document`.
    ///
    /// Setting the position of a table will only affect output when
    /// `Document::to_string_in_original_order` is used.
    pub fn set_position(&mut self, position: usize) {
```

and 

```rs
    /// The position of the `Table` within the `Document`.
    ///
    /// Returns `None` if the `Table` was created manually (i.e. not via parsing)
    /// in which case its position is set automatically.
    pub fn position(&self) -> Option<usize> {
        self.position
    }
```

that made me think the position is exclusive to parsed elements.

But it seems to work with something like this 🚀

```rs
                let mut profile = toml_edit::Table::new();
                profile.set_implicit(true);
                profile.insert("release", toml_edit::Item::Table(toml_edit::Table::new()));
                profile.set_position(3);
                doc.insert("profile", toml_edit::Item::Table(profile));
```

If I insert a new Table into a document, do I have to manually adjust the positions of all elements later on?

epage:
> If I insert a new Table into a document, do I have to manually adjust the positions of all elements later on?

Yes.  Something we could do to improve this is to sort by `(pos, path)` and not just `pos`.  This would allow people to reuse an existing tables position and get more predictable results.

[...]

PR #292, opened 2022-01-07
docs: Clarify table position

epage:
Inspired by #291

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -10,15 +10,16 @@ use crate::{InlineTable, InternalString, Item, KeyMut, Value};
 /// Type representing a TOML non-inline table
 #[derive(Clone, Debug, Default)]
 pub struct Table {
-    // comments/spaces before and after the header
+    // Comments/spaces before and after the header
     pub(crate) decor: Decor,
-    // whether to hide an empty table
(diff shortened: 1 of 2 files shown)


src/table.rs
```diff
@@ -10,15 +10,16 @@ use crate::{InlineTable, InternalString, Item, KeyMut, Value};
...
-    // whether this is a proxy for dotted keys
+    // Whether this is a proxy for dotted keys
     pub(crate) dotted: bool,
-    // used for putting tables back in their original order when serialising.
-    // Will be None when the Table wasn't parsed from a file.
-    pub(crate) position: Option<usize>,
+    // Used for putting tables back in their original order when serialising.
+    //
+    // `None` for user created tables (can be overridden with `set_position`)
+    doc_position: Option<usize>,
```
webmaster128, line 22: Is this position always in a `Document`? Or is this used for nested tables as well? Would `parent_position` as in "position in the parent" be a better name?

epage: No matter the nesting level, its is the global position in the document or otherwise we wouldn't be able to preserve cases like
```toml
[foo.bar]

[baz]

[foo.bar.yum]

[foo]
```

PR #304, opened 2022-01-26
Slightly improve `on_comment` performance

mati865:
On my PC this is very slight improvement for `cargo` benchmark, `linear` didn't show measurable difference.
Tested on x86_64 Linux.

### DIFF
--- a/src/parser/document.rs
+++ b/src/parser/document.rs
@@ -134,8 +134,7 @@ impl TomlParser {
     }
 
     fn on_comment(&mut self, c: &str, e: &str) {
-        self.trailing.push_str(c);
-        self.trailing.push_str(e);
+        self.trailing = [&self.trailing, c, e].concat();
     }
 
     fn on_keyval(&mut self, mut path: Vec<Key>, mut kv: TableKeyValue) -> Result<(), CustomError> {


epage (approved):

ordian:
Interesting. I'm curious to look at the generated code and see if the difference is of statistical significance.

epage:
My guess is that its saving an allocation.  I'm assuming `concat` looks up all of the sizes and does a single allocation rather than two.

epage:
FYI the thing making me hesitate to hit the merge button is reminding myself for whether a straight `self = trailing` is safe, if there is a case where we might overwrite data.  I'd need time to dig back into the parser to refresh myself on that point.

btw If it is the allocations, another way of addressing this is
```rust
self.trailing.reserve(c.len() + e.len());
self.trailing.push_str(c);
self.trailing.push_str(e);
```

ordian:
Yeah, sounds plausible https://rust.godbolt.org/z/e5WfbME79. However, running `cargo bench cargo` is noisy on my laptop and didn't show any difference other than noise.

> whether a straight self = trailing is safe, if there is a case where we might overwrite data. I'd need time to dig back into the parser to refresh myself on that point.

sorry, I don't follow

epage:
> whether a straight self = trailing is safe, if there is a case where we might overwrite data. I'd need time to dig back into the parser to refresh myself on that point.

The code change is from
```rust
self.trailing.push_str(c);
self.trailing.push_str(e);
```
to
```rust
self.trailing = [&self.trailing, c, e].concat();
```

The first appends while the second one overwrites.  If it is always intended to be empty before the overwrite, then this is safe.  If it isn't, then we lose data.  I've not taken time to double check this.  I'm assuming its safe, that we'll only hit this after a table `take`s it.

ordian:
Still don't follow, sorry. The append to self is equivalent to overwriting self with (self + appended).

ordian:
I think we're spending too much time on this change anyway. Thanks @mati865

mati865:
```rust
self.trailing.reserve(c.len() + e.len());
self.trailing.push_str(c);
self.trailing.push_str(e);
```
I've tried that first and to my surprise it was small but consistent slowdown for me (haven't investigated why).

> whether a straight self = trailing is safe, if there is a case where we might overwrite data. I'd need time to dig back into the parser to refresh myself on that point.

IMO if safe code like that could cause data loss we all would be in troubles.

> Yeah, sounds plausible rust.godbolt.org/z/e5WfbME79.

Yup, it's nicely inlined.

> However, running cargo bench cargo is noisy on my laptop and didn't show any difference other than noise.

Yeah, those benchmarks are very noisy but across multiple runs this is at least 1% improvement for me.

Attaching small shots of flamegraph to illustrate the difference:
- old: 
obraz

- new: 
obraz

> Thanks @mati865

Thank you for quick review.

epage:
> self.trailing = [&self.trailing, c, e].concat();

Somehow I kept overlooking that this was three elements, not 2.

Issue #340, opened 2022-09-23
Migrate `toml` to be on `toml_edit`


epage:
`toml` will be a wrapper around the `toml_edit` crate, providing serde integration, decoupled from the format-preserving parsing, and offering a more stable API

As part of switching cargo adopting `toml_edit`, `toml_edit::easy`, a copy of `toml`s API, was created so cargo could have a single toml parser for both serde and editing.

We plan to move `toml_edit::easy` into the `toml` crate, providing nearly the same API at nearly the same performance while reducing overall maintenance effort and allowing people who have both needs to have a common parser.

Blockers
- [x] #327
- [x] #302
- [x] #303
- [ ] Release a 0.5 with deprecations and 1.48 MSRV (https://github.com/toml-rs/toml/issues/346)
  - `Deserializer::end`
  - `Deserializer::set_require_newline_after_table`
  - `Deserializer::set_allow_duplicate_after_longer_table`

Performance
- See https://github.com/rust-lang/cargo/pull/10086#issuecomment-1006751490

Compatibility
- https://github.com/toml-rs/toml/issues/287
- https://github.com/toml-rs/toml/pull/344

epage:
@samuelcolvin 

> How does `toml_edit`s performance compare to `toml-rs`?

Last time we looked at this was when deciding if cargo could switch to toml_edit and we found micro-benchmarks to show toml_edit was up to 5% slower.  See https://github.com/rust-lang/cargo/pull/10086#issuecomment-1006751490

samuelcolvin:
@epage I don't understand why you've closed all the issues and pull requests on toml-rs? 

If the library is no longer maintained, mark it as such, but just closing all issues makes it harder to understand what issues toml-rs had, and harder to maintain it going forward. 

You've effectively just destroyed the fastest rust toml parser.

I don't think this is a sensible or considerate way to "maintain" an open source library that many people rely on.

It would be bad enough if you were the sole or primary author, but you're not afaik.

Quite disappointed by this 😞.

epage:
[...]

PR #347, opened 2022-09-24
Make UnsupportedType error more informative

mgsloan:
Moved this over from a PR on the other repo https://github.com/toml-rs/toml-rs/pull/369

That PR didn't get merged in 3 years due to no other breaking changes to batch it with.  @alexcrichton said:

> Indeed yes, still interested in merging, but holding off until this crate has a breaking release, which it currently isn't scheduled yet to have.

UnsupportedType with no further info is not very helpful when trying to serialize a complex structure. This adds more information so that the source of the problem is clearer to the developer.

I realize that this is a breaking change. An alternative I considered was adding a variant to the enum, but I think this is better. Cases where users are explicitly matching on this probably ought to be updated - hopefully there are few in the wild.

### DIFF
--- a/crates/toml/src/ser.rs
+++ b/crates/toml/src/ser.rs
@@ -113,13 +113,20 @@ pub enum Error {
     ///
     /// Currently the TOML format does not support serializing types such as
     /// enums, tuples and tuple structs.
-    UnsupportedType,
+    UnsupportedType {
+        /// Name of the type for which serialization is not supported, if any.
+        name: Option<String>,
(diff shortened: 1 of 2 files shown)


epage:
https://github.com/toml-rs/toml/issues/340 will include a breaking change.  That'd be a good time to pull this in.

epage:
#464 improved this to a degree and I'm unsure of the value of including the extra information, so closing this out for now.  Let's discuss further needs in a new issue.

Issue #326, opened 2022-07-23
feature request: Deserialize from `Read` and Serialize into `Write`


NobodyXu:
`toml_edit` currently requires the entire file to be read into `Vec<u8>` to deserialize and requires an intermediate `Vec<u8>` to serialize.

Having APIs for deserializing from `Read` and serialize directly into `Write` would avoid the intermediate buffer, provides better performance and less memory usage.

epage:
So this isn't just supporting `Read` and `Write` but doing incremental parsing while doing using those traits.

Could you describe your use case where loading the toml into a buffer isn't sufficient?  Incremental parsing like that adds a lot of complexity and its hard to see the justification for it.

NobodyXu:
@epage Not sure whether "incremental parsing" is the right word, but I just want something like [`serde_json::from_reader`](https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html) and [`serde_json::to_writer`](https://docs.rs/serde_json/latest/serde_json/fn.to_writer.html) where serializing/deserializing can be done without reading the entire file/writing the entire serialized content into `Vec<u8>`.

Looking at the implementation of `serde_json::Serialize`, this doesn't seem to be that hard and actually `serde` now recommends new format implementation to accept `Read` and `Write` and only provides a convenient function for reading from slice and writing to `Vec<u8>`.

epage:
> this doesn't seem to be that hard

The parser has to be written with this use case in mind and can also potentially negatively impact performance to still get all of the desired feature (the error message with line/column information).  We have a very tight performance budget so we can be used in crates like cargo

epage:
I'd also like to re-iterate my question

> Could you describe your use case where loading the toml into a buffer isn't sufficient? 

Generally TOML files are meant to be human edited which means their size is more constrained.  The use case for a TOML file where it is parsed a chunk at a time rather than loaded upfront seems relatively small.  We'd need to understand the use cases to determine if the benefits outweigh the costs.

NobodyXu:
> Could you describe your use case where loading the toml into a buffer isn't sufficient?

I need to update `~/.cargo/.crates.toml` and I have no control on its size.

epage:
While in theory that file can be unbounded, are you aware of any cases where it exceeds 1000 entries?

NobodyXu:
> While in theory that file can be unbounded, are you aware of any cases where it exceeds 1000 entries?

No, at least not on my computer.

epage:
[...]

PR #349, opened 2022-09-26
Add from_reader and to_writer methods

WhyNotHugo:
These match what other serde serialisers offer.

This is a rebase of https://github.com/toml-rs/toml-rs/pull/452 which was trying to resolve https://github.com/toml-rs/toml-rs/issues/215, resolves https://github.com/toml-rs/toml-rs/issues/216.

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -6,9 +6,11 @@
 
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::convert::From;
 use std::error;
 use std::f64;
 use std::fmt;
+use std::io;
 use std::iter;
 use std::marker::PhantomData;
 use std::str;
@@ -80,6 +82,46 @@ where
     Ok(ret)
 }
 
+/// Deserializes an IO stream into a type.
+///
+/// This function will attempt to read a TOML document from `r` and
+/// deserialize to `T` .
+///
+/// # Examples
+///
+/// ```rust
+/// #[macro_use]
+/// extern crate serde_derive;
+/// extern crate toml;
+///
+/// use std::fs::File;
+///
+/// #[derive(Debug, Deserialize)]
+/// struct Config {
+///     title: String,
+///     owner: Owner,
+/// }
+///
+/// #[derive(Debug, Deserialize)]
+/// struct Owner {
+///     name: String,
+/// }
+///
+/// let mut file = File::open("config.toml").unwrap();
+/// let config: Config = toml::from_reader(&mut file).unwrap();
+///
+/// println!("{:?}", config);
+/// ```
+pub fn from_reader<R, T>(r: &mut R) -> Result<T, ReaderError>
+where
+    R: io::Read,
+    T: de::DeserializeOwned,
+{
+    let mut buf = Vec::new();
+    r.read_to_end(&mut buf)?;
+    from_slice::<T>(&buf).map_err(ReaderError::ParserError)
+}
+
 /// Errors that can occur when deserializing a type.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct Error {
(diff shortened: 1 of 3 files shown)


crates/toml/src/de.rs
```diff
@@ -96,6 +142,21 @@ struct ErrorInner {
     key: Vec<String>,
 }
 
+/// Error that can occur when deserializing via IO.
+#[derive(Debug)]
+pub enum ReaderError{
```
epage, line 147: How come this is using a custom error type?

crates/toml/src/ser.rs
```diff
@@ -91,6 +92,60 @@ where
...
+///             connection_max: 5000,
+///             enabled: false,
+///         },
+///     };
+///
+///     let mut file = File::create("config.toml").unwrap();
+///     toml::to_writer(&mut file, &config).unwrap();
+/// }
+/// ```
+pub fn to_writer<W, T: ?Sized>(writer: &mut W, value: &T) -> Result<(), Error>
```
epage, line 139: I see that `serde_json` offers a `to_writer_pretty`, should we do the same?

epage (commented):
Thanks!

crates/toml/src/de.rs
```diff
@@ -80,6 +82,50 @@ where
...
+///
+/// This function will attempt to read a TOML document from `r` and
+/// deserialize to `T` .
+///
+/// # Examples
+///
+/// ```rust
+/// #[macro_use]
+/// extern crate serde_derive;
+/// extern crate toml;
```
epage, line 95: Could you drop the use of `extern crate`?

crates/toml/src/de.rs
```diff
@@ -80,6 +82,50 @@ where
...
+/// }
+///
+/// #[derive(Debug, Deserialize)]
+/// struct Owner {
+///     name: String,
+/// }
+///
+/// fn main() {
+/// # }
+/// # fn fake_main() {
```
epage, line 112: Let's try to avoid having a `main()`

crates/toml/src/ser.rs
WhyNotHugo: There's no minified toml, is there? It's always "pretty".

crates/toml/src/de.rs
epage: Please come back with a proposal on that.  I would prefer to keep to a single error type especially because others, at least serde_json,. get away with one error type.

crates/toml/src/ser.rs
epage: There are different ways to format the output.  See https://docs.rs/toml/latest/toml/ser/struct.Serializer.html#method.pretty

crates/toml/src/de.rs
WhyNotHugo: Using the same error for both doesn't really seem to make sense TBH. For example:

- When serialising, there's no `line` and `column` related to the error, but this must be allocated and returned in the error anyway.
- The [`Error.category`](https://docs.rs/serde_json/latest/serde_json/error/enum.Category.html) has categories that don't apply to both operations, but application code using the library needs to handle both. When serialising, `Category::Syntax` does not make sense as a possible return value, but needs to be handled anyway. Of course, it's possible for usages to simply use `Category::Syntax => unreachable!()`, but that's an indicator of a poor API.

WhyNotHugo: I guess we could use `line: Option<usize>` and use more generic categories. Like `Category::InvalidData` when either the `JSON` input is invalid, or when the input data cannot be serialised into json.

sammyne:
Any plan of merging this PR？@epage

WhyNotHugo:
@sammyne This needs to be rebased and the errors redesigned as mentioned above.

epage:
This needs more discussion (see #538), moving it to #326

WhyNotHugo:
Thanks for picking up on this and pushing it through :)

PR #353, opened 2022-10-03
docs: Update fuzz readme

manunio:
Hi, Running `cargo +nightly fuzz run parse_document` will error out with:
```
Error: could not read the manifest file: toml/fuzz/Cargo.toml

Caused by:
    No such file or directory (os error 2)
```

As a solution, Running `cargo +nightly fuzz run parse_document --fuzz-dir=.` will point it to toml_edit_fuzz manifest file.


epage:
I've not really looked much into this.  I take it `cargo fuzz` has an expectation for what the directory heirarchy looks like?  I was just trying to keep the workspace more flag but we can consider changing that as well.

PR #357, opened 2022-10-10
Fix typo in docs

clint-white:
The space following the first asterisk meant that it was rendered as a literal asterisk instead of opening an emphasis region.

I also wonder about another typo in that sentence: was it intended to read "of items" instead of "or items"?  I wasn't sure about that one, so I left it unchanged for now.


epage:
Thanks!

PR #372, opened 2022-11-23
chore(gh): add CIFuzz action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where toml(_edit?) already runs (https://github.com/google/oss-fuzz/tree/master/projects/toml_edit). CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'toml_edit'
+        dry-run: false
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'toml_edit'
+        fuzz-seconds: 300
```
epage, line 18: With having a cap on the maximum number of parallel runners, I feel like 300 seconds is too much and would negatively affect CI times

DavidKorczynski: Apologies for the delay @epage -- I'm happy to address this: would would your preference be? We could go down to 15 seconds?

epage: 15s shouldn't be a problem

DavidKorczynski: Sounds good! Fixed it and updated to the lastest CIFuzz action as well

PR #398, opened 2022-12-20
allows for empty `toml::toml!` invocations

sanbox-irl:
I don't know too much about macros, so perhaps this is too noisy a change, but this simply allows calling `toml::toml!{}` without anything inside it, creating a blank `toml::Value::Table`. 

In unit tests, it can read more clearly for an empty toml document when other tests nearby are constructing real data, and it has a very small footprint, so it doesn't seem like an issue in general.

### DIFF
--- a/crates/toml/src/macros.rs
+++ b/crates/toml/src/macros.rs
@@ -28,10 +28,13 @@ use crate::value::{Array, Table, Value};
 /// ```
 #[macro_export]
 macro_rules! toml {
-    ($($toml:tt)+) => {{
+    ($($toml:tt)*) => {{
         let table = $crate::value::Table::new();
+
+        #[allow(unused_mut)]
         let mut root = $crate::Value::Table(table);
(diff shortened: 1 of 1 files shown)


crates/toml/src/macros.rs
```diff
@@ -28,10 +28,13 @@ use crate::value::{Array, Table, Value};
 /// ```
 #[macro_export]
 macro_rules! toml {
-    ($($toml:tt)+) => {{
+    ($($toml:tt)*) => {{
```
epage, line 31: Could you add a test case for empty tables?

epage (commented):
As a heads up, I'm trying to hold off on releasing 0.6 (master) without some specific changes.  What we can do, once this is merged, is you can submit another PR that cherry-picks this commit to the `toml-v0.5` branch and I can release that.

sanbox-irl:
sounds great -- not sure i'm following re: cherry picking, but that's a future problem, for the future

PR #399, opened 2022-12-21
refactor(toml): clippy

azzamsa:


crates/toml/src/de.rs
```diff
@@ -518,7 +518,7 @@ impl<'de, 'b> de::MapAccess<'de> for MapVisitor<'de, 'b> {
         let res = seed.deserialize(MapVisitor {
             values: Vec::new().into_iter().peekable(),
             next_value: None,
-            depth: self.depth + if array { 0 } else { 1 },
+            depth: self.depth + usize::from(!array),
```
epage, line 521: I actually disagree with this clippy lint and would rather have it enabled across the entire lib.

(the default-enabled clippy lints are controversial enough that it isn't used in the `cargo` repo)

azzamsa: Is this okay?

```diff
modified   crates/toml/src/de.rs
@@ -518,7 +518,7 @@ impl<'de, 'b> de::MapAccess<'de> for MapVisitor<'de, 'b> {
         let res = seed.deserialize(MapVisitor {
             values: Vec::new().into_iter().peekable(),
             next_value: None,
-            depth: self.depth + usize::from(!array),
+            depth: self.depth + if array { 0 } else { 1 },
             cur_parent: self.cur - 1,
             cur: 0,
             max: self.max,
modified   crates/toml/src/lib.rs
@@ -137,6 +137,7 @@
 //! [`serde`]: https://serde.rs/
 //! [serde]: https://serde.rs/
 
+#![allow(clippy::bool_to_int_with_if)]
 #![deny(missing_docs)]
 #![warn(rust_2018_idioms)]
 // Makes rustc abort compilation if there are any unsafe blocks in the crate.
 ```

epage: Yup

epage:
Hmm, this is a pednatic clippy lint which we already have marked as `allow`, so unsure why this is needed.  Its also stale because I hadn't gotten back to this.

PR #400, opened 2022-12-21
feat(toml): add field `message` for `de::Error`

azzamsa:
It produces the cause of the error without additional line and column information.

--- 

I want to get the field `message` from `ErrorInner`.  Unfortunately, there trivial way to do it but to add a `message()` function.

Using `Display` for `de::Error` I got additional line and column information. In the case of a missing field, this info is incorrect.  It always says the missing field is in `line 1` and  `column 1` (I suspect this is the default). Because I can put the missing field anywhere in the toml file and it works.

 ```rust
 use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct Config {
    global_string: String,
    global_integer: i32,
}

fn main() {
    let input: &str = r#"
        # global_string = "test"
        global_integer = 5
    "#;

    match toml::from_str::<Config>(input) {
        Ok(decoded) => {
            println!("{:#?}", decoded);
        }
        Err(e) => {
            // dbg!(&e.line_col());
            dbg!(&e);
            eprintln!("{}", e);
        }
    };
}
```

```rust
     Running `target/debug/playground`
[src/main.rs:21] &e = Error {
    inner: ErrorInner {
        kind: Custom,
        line: Some(
            0,
        ),
        col: 0,
        at: Some(
            0,
        ),
        message: "missing field `global_string`",
        key: [],
    },
}
missing field `global_string` at line 1 column 1
```

I don't want to confuse my user. I just want to use the message from the message field.

I use `message` here for the function name. If you are okay with this change. Please suggest me a better function name and docstring.

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -2016,6 +2016,11 @@ impl Error {
         self.inner.line.map(|line| (line, self.inner.col))
     }
 
+    /// Produces the cause of the error wihout additional line and colum information
+    pub fn message(&self) -> String {
+        self.inner.message.to_owned()
+    }
+
     fn from_kind(at: Option<usize>, kind: ErrorKind) -> Error {
         Error {
             inner: Box::new(ErrorInner {


epage:
I'd like to hold off on changes like this until #340 as that will dramatically change things and all of the work will need to be re-implemented.

epage:
Also, I would encourage first opening an issue for us to discuss needs and alignment before opening a PR.

epage:
> Oh, I see. So I don't waste my time if the idea is rejected in the first hand. Thanks.

Yes, the idea is to help avoid wasting time but it isn't just for rejecting things.  To clarify, there are
- Issues to resolve on the problem (discussed in Issues)
- Issues to resolve on the proposed solution (discussed in Issues)
- Issues to work out with the implementation (discussed in PR)

There is also a social factor to this.  Maintainers usually have enough empathy to be concerned with wasting contributors time and discouraging them.  Starting with a PR then puts a social pressure on a maintainer which can increase the workload and stress on them.

azzamsa:
Very thorough points. Just realize about the social pressure things. Thanks a lot!

epage:
This is superseded by #473

PR #412, opened 2022-12-30
perf(parser): Reduce error type's size

epage:
This ends up being a wash or worse performance
- Success route is about 6 pointers
- Failure was about 8 pointers but this dropped it to 7.

Overall, we'd need more reduction in size on both branches before we could see performance improvements.


epage:
Similarly, when introducing `nom8::input::Located`, I tried to put the location in the error, rather than the pointer, and it similarly made things worse.  I also tried `IntoOutput::into_outout` without gains

Issue #287, opened 2021-12-31
Rendering of strings doesn't match toml-rs


epage:
While we intentionally chose to prefer `"` over `'` when no escaping is needed (because this is most likely what users will do), we should probably use `"""` to avoid quoting `"` in strings.

We also need to look into our `\t` handling.

Compare
- [toml-rs pretty_tricky](https://github.com/alexcrichton/toml-rs/blob/master/test-suite/tests/pretty.rs#L178)
- Our pretty_tricky added in https://github.com/ordian/toml_edit/pull/286

epage:
See also #752

epage:
At this point, this likely doesn't matter

PR #470, opened 2023-01-19
feat(toml)!: Update generator by using `toml_edit`

### LINKED ISSUE #396: More `Error(UnsupportedType)` Feedback
Really helpful library, thanks so much for your work on it. 

This issue is borne of https://github.com/dan-fritchman/Layout21/issues/33, in which: 

* I have a big tree of `struct`s, with lots of fields, internal types, and external types along the way
* I pass the biggest top-most one to `toml::to_string()`
* What comes back is this:

```
failures:

---- tests::empty_lib_to_toml stdout ----
thread 'tests::empty_lib_to_toml' panicked at 'save failed: Error(UnsupportedType)', gds21/src/tests.rs:214:50
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Question: is there any good way to get more than `UnsupportedType` as feedback here?  
(Set a setting, submit a PR to improve the error message, whatever.) 

Thanks again!

epage:
This is the other main half of #340.  Still have deprecations and tests left

Note that strings are rendered differently, see #287

By extension this also finishes up #396.

BREAKING CHANGES
- `impl Display for toml::Value` now renders as values, not documents, see instead `Table`
- `toml::ser::Serializer` only serializes documents, instead see `toml::ser::ValueSerializer`
- `toml::ser::tables_last` is removed, no longer needed
- `toml::ser::to_vec` is removed to mirror the loss of `from_slice`
- atm `toml::ser::to_string_pretty` just causes larger arrays to be indented
- `toml::ser::Error` is now opaque
- `toml::ser::Serializer::pretty_string` was removed
- `toml::ser::Serializer::pretty_string_literal` was removed
- `toml::ser::Serializer::pretty_array` was removed
- `toml::ser::Serializer::pretty_array_indent` was removed
- `toml::ser::Serializer::pretty_array_trailing_comma` was removed
- `toml::ser::Serializer` is now used used by value, rather than `&mut`

Fixes #396

### DIFF
--- a/crates/toml/src/fmt.rs
+++ b/crates/toml/src/fmt.rs
@@ -1,0 +1,68 @@
+#[derive(Copy, Clone)]
+pub(crate) struct DocumentFormatter {
+    pub(crate) multiline_array: bool,
+}
+
+impl Default for DocumentFormatter {
+    fn default() -> Self {
+        Self {
+            multiline_array: false,
+        }
+    }
+}
+
+impl toml_edit::visit_mut::VisitMut for DocumentFormatter {
+    fn visit_document_mut(&mut self, node: &mut toml_edit::Document) {
+        toml_edit::visit_mut::visit_document_mut(self, node);
+    }
+
+    fn visit_item_mut(&mut self, node: &mut toml_edit::Item) {
+        let other = std::mem::take(node);
+        let other = match other.into_table().map(toml_edit::Item::Table) {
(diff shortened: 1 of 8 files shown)


danieleades:
what should be used instead of `toml::to_vec`?

epage:
`toml::to_string(&data).into_bytes()`

PR #498, opened 2023-01-27
Turn off serde "derive" feature

dtolnay:
I think this is an unreasonably heavy dependency for this one trivial impl, especially in a module that already contains a 147-line bespoke Deserialize impl.


epage:
*shrug* It was going away soon anyways but sure

PR #502, opened 2023-01-30
display feature guarded items on docs.rs

robjtede:
E.g., currently, the serde impls for `Spanned` are not shown at all on docs.rs.

Added visibility for all publicly visible items with feature guarded blocks in `serde-spanned` and `toml-datetime`. Will do the other crates if this is accepted.

### DIFF
--- a/crates/toml_datetime/src/lib.rs
+++ b/crates/toml_datetime/src/lib.rs
@@ -9,6 +9,7 @@
 // and lets them ensure that there is indeed no unsafe code as opposed to
 // something they couldn't detect (e.g. unsafe added via macro expansion, etc).
 #![forbid(unsafe_code)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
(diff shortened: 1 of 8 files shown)


crates/toml_datetime/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 // and lets them ensure that there is indeed no unsafe code as opposed to
 // something they couldn't detect (e.g. unsafe added via macro expansion, etc).
 #![forbid(unsafe_code)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
epage, line 12: Could you instead switch these to ` #![cfg_attr(docsrs, feature(doc_auto_cfg))] ` and remove the `doc(cfg(...))` attributes?

robjtede: i'm staying away from it personally until it's stable but sure, updated

PR #505, opened 2023-01-31
feat: support deserializing types that borrow data in `from_str`

AlexTMjugador:
Some structs users may define, such as the following one, [do not implement serde's `DeserializeOwned` trait because they might borrow data from the deserializer](https://serde.rs/lifetimes.html), and thus they can't be deserialized for any lifetime:

```rust
#[derive(Deserialize)]
struct Message<'a> {
    #[serde(borrow)]
    title: Cow<'a, str>,
    #[serde(borrow)]
    text: Cow<'a, str>
}
```

Even though, as far as I know, the TOML deserializer implemented in these crates doesn't take advantage of zero-copy deserialization features, the fact that the convenience `from_str` function has a `DeserializeOwned` bound prevents it from working with such structs. [This is a regression from previous versions of the crate](https://github.com/toml-rs/toml/blob/99332f773dcf3ba3f627d842b4f9116b0b7192ff/crates/toml/src/de.rs#L73-L81).

For comparison, [`serde_json`'s equivalent function has a `Deserialize<'a>` bound instead](https://docs.rs/serde_json/1.0.91/serde_json/fn.from_str.html), where `'a` is the lifetime of the input string, that supports these structs.

Replacing `toml::from_str` with the more verbose expression `T::deserialize(toml::de::Deserializer::new(s))` works in these cases, so I think there is no compelling reason to make this function more restrictive than it needs to be.

This change should be semver-compatible according to my best interpretation of the [API evolution RFC](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md).

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -38,9 +38,9 @@
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
 #[cfg(feature = "parse")]
-pub fn from_str<T>(s: &'_ str) -> Result<T, Error>
+pub fn from_str<'a, T>(s: &'a str) -> Result<T, Error>
 where
-    T: serde::de::DeserializeOwned,
+    T: serde::de::Deserialize<'a>,
 {
     T::deserialize(Deserializer::new(s))
 }
--- a/crates/toml/tests/testsuite/enum_external_deserialize.rs
+++ b/crates/toml/tests/testsuite/enum_external_deserialize.rs
@@ -23,9 +23,9 @@ struct Multi {
     enums: Vec<TheEnum>,
 }
 
-fn value_from_str<T>(s: &'_ str) -> Result<T, toml::de::Error>
+fn value_from_str<'a, T>(s: &'a str) -> Result<T, toml::de::Error>
 where
-    T: serde::de::DeserializeOwned,
+    T: serde::de::Deserialize<'a>,
 {
     T::deserialize(toml::de::ValueDeserializer::new(s))
 }
--- a/crates/toml/tests/testsuite/serde.rs
+++ b/crates/toml/tests/testsuite/serde.rs
@@ -1064,6 +1064,7 @@ fn datetime_offset_issue_496() {
 fn borrowed_data() {
     #[derive(Serialize, Deserialize)]
     struct Foo<'a> {
+        #[serde(borrow)]
         a: Cow<'a, str>,
     }
 
--- a/crates/toml/tests/testsuite/serde.rs
+++ b/crates/toml/tests/testsuite/serde.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use serde::Deserialize;
 use serde::Deserializer;
 use serde::Serialize;
@@ -1057,4 +1058,16 @@ fn datetime_offset_issue_496() {
     let toml = original.parse::<toml::Table>().unwrap();
     let output = toml.to_string();
     snapbox::assert_eq(original, output);
+}
+
+#[test]
+fn borrowed_data() {
+    #[derive(Serialize, Deserialize)]
+    struct Foo<'a> {
+        a: Cow<'a, str>
+    }
+
+    let toml = map! { a: Value::String(String::from("bar")) };
+    assert!(toml.clone().try_into::<Foo>().is_ok());
+    assert!(toml::from_str::<Foo>(&toml.to_string()).is_ok());
 }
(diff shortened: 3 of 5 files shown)


crates/toml/src/de.rs
```diff
@@ -38,9 +38,9 @@
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
 #[cfg(feature = "parse")]
-pub fn from_str<T>(s: &'_ str) -> Result<T, Error>
+pub fn from_str<'a, T>(s: &'a str) -> Result<T, Error>
 where
-    T: serde::de::DeserializeOwned,
+    T: serde::de::Deserialize<'a>,
```
epage, line 43: > This change should be semver-compatible according to my best interpretation of the [API evolution RFC](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md).

This changes the trait bounds from a more general one to a more specific one which is a breaking change

crates/toml/src/de.rs
```diff
@@ -38,9 +38,9 @@
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
 #[cfg(feature = "parse")]
-pub fn from_str<T>(s: &'_ str) -> Result<T, Error>
+pub fn from_str<'a, T>(s: &'a str) -> Result<T, Error>
```
epage, line 41: Please add a test for borrowed data to ensure this works.

If it does work, then more investigation is needed because it shouldn't [according to the serde docs](https://serde.rs/lifetimes.html#trait-bounds).  `&str` gets parsed into owned types and then we deserialize that.

epage, line 41: This change should be reflected in `toml_edit` as well

crates/toml/src/de.rs
AlexTMjugador: Thanks for the review! :smile: Isn't it the reverse, though? [`DeserializeOwned` is syntax sugar for the higher-ranked trait bound `for<'de> Deserialize<'de>`](https://docs.rs/serde/1.0.152/serde/de/trait.DeserializeOwned.html), so I'd argue that `DeserializeOwned` is more specific. The [HRTB's RFC seems to support this](https://rust-lang.github.io/rfcs/0387-higher-ranked-trait-bounds.html#subtyping-of-trait-references), stating that:

> [...] if I have a trait reference like `for<'a> FnMut(&'a int)`, that would be usable wherever a trait reference with a concrete lifetime, like `FnMut(&'static int)`, is expected.

crates/toml/src/de.rs
AlexTMjugador: I've just added the test in https://github.com/toml-rs/toml/pull/505/commits/44563c2879b8d1b73caa17b9e43d216e3737948e, thanks!

I think that the tested code works because `String`s, being owned types, have a implicit `'static` lifetime bound (they can live indefinitely long until dropped). `Cow`s support storing owned data, so a valid lifetime for a `Cow` that either stores owned data or static strings is `Cow<'static, str>`.  Due to [lifetime coercion](https://doc.rust-lang.org/rust-by-example/scope/lifetime/lifetime_coercion.html), a `'static` lifetime can be used whenever a shorter lifetime is expected, such as `'de` in this case: the lifetime of the serialized data source.

epage: Is that sufficient if the user explicitly declared the field as `borrow`?

What about if the user isn't using `Cow`?

AlexTMjugador: Done!

epage: First, this isn't a morality discussion.

Second, we are shifting a compile time error into a runtime error for the easy path while, as you pointed out, a workaround is available for those that want to borrow.  In API design, we should be emphasizing the safe path with the easy path while ideally still supporting the unsafe route.  The current solutions fits that to a T.  I don't know why the serde docs make the recommendation they do but in this case, I see it being useful.

I'm going to go ahead and close this and ask any further discussion to happen in the appropriate issue (#490).

epage:
From https://github.com/toml-rs/toml/pull/505#discussion_r1092513171

> Second, we are shifting a compile time error into a runtime error for the easy path while, as you pointed out, a workaround is available for those that want to borrow. In API design, we should be emphasizing the safe path with the easy path while ideally still supporting the unsafe route. The current solutions fits that to a T. I don't know why the serde docs make the recommendation they do but in this case, I see it being useful.
> 
> I'm going to go ahead and close this and ask any further discussion to happen in the appropriate issue (#490).

PR #508, opened 2023-02-02
Fix crate version in example

matthiasbeyer:
Found this while browsing the docs on docs.rs.

### DIFF
--- a/crates/toml/src/lib.rs
+++ b/crates/toml/src/lib.rs
@@ -5,7 +5,7 @@
 //! ```toml
 //! [package]
 //! name = "toml"
-//! version = "0.4.2"
+//! version = "0.7.1"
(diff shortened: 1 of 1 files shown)


crates/toml/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! ```toml
 //! [package]
 //! name = "toml"
-//! version = "0.4.2"
+//! version = "0.7.1"
 //! authors = ["Alex Crichton <alex@alexcrichton.com>"]
 //!
 //! [dependencies]
```
epage, line 11: This is meant to be a real world example of TOML.  I do not see the value in creating churn in updating the version on every release though.

matthiasbeyer: Okay :smiley:

PR #536, opened 2023-03-30
serde_spanned: fix copypasta in crate description

decathorpe:
It looks like the crate description for the `serde_spanned` crate was copy-pasted from the `toml` crate. I've replaced it with the heading from the crate's docs instead.


epage:
Thanks!

PR #538, opened 2023-04-02
Fixed spelling errors

Cactus-man:
"if" was spelled "iff" in doc comments in `toml_edit/src/table.rs`


crates/toml_edit/src/table.rs
```diff
@@ -265,7 +265,7 @@ impl Table {
         self.items.iter().filter(|i| !(i.1).value.is_none()).count()
     }
 
-    /// Returns true iff the table is empty.
+    /// Returns true if the table is empty.
```
epage, line 268: `iff` is notation for "if and only if" but "if" also works here.

Issue #544, opened 2023-04-13
Provide an easy way to deserialize the `toml_edit::Item`


SomeoneToIgnore:
See https://github.com/toml-rs/toml/pull/543#issuecomment-1506943136

In my application, I have a toml-based config that is getting assembled out of multiple toml pieces and overrides.
I end up having a `toml::Item` constructed in runtime and a need to deserialize it into internal struct.

Before, I could've used `toml_edit::de::from_item` which is now removed and `ValueDeserializer` is provided as an alternative.
Yet, the only possible way to construct this is to convert the `Item` into a `String`, then parse this string into `ValueDeserializer`?
In addition to that, table `Item`s have to be somehow created as inline table form, since the deserializer complains when parsing a regular form.

I propose to make `ValueDeserializer::new` constructor public — if a person has a correct `Item` on hands, it should be able to directly covert it into any `T: DeserializeOwned` or similar.

SomeoneToIgnore:
Here's an example of dynamic construction and further parsing: https://github.com/neondatabase/neon/blob/fd31fafeeeb46ab1f6a68d888dd9b1bf2c1db816/pageserver/src/config.rs#L728

Some of the values per key could be replaced with the overrides, that we're sure is a valid toml, but then we need to parse these possibly-replaced items and `ValueDeserializer` does not help much for table case.

epage:
If you have an `Item`, you can call [`Item::into_value`](https://docs.rs/toml_edit/latest/toml_edit/enum.Item.html#method.into_value) and then use [`IntoDeserializer`](https://docs.rs/toml_edit/latest/toml_edit/enum.Value.html#impl-IntoDeserializer%3C'de,+Error%3E-for-Value) to get `ValueDeserializer`.

The shift in focus away from `Item` was an intentional change due to issues I found in `toml` where it conflated TOML documents and TOML values.  As a result of that, we shifted focus towards Document, Table, and Value as "top level" types to be interacted with.

At the moment, the API focuses on the core workflow and just provides the building blocks for other workflows.  This means that there are extra hoops to go through, like in this case.  As we get input on what people are doing, we can consider re-evaluating some of these.

SomeoneToIgnore:
Thank you so much, that chain of conversions indeed did the trick.
[...]

PR #545, opened 2023-04-14
Update ValueDeserializer creation docs

SomeoneToIgnore:
Closes https://github.com/toml-rs/toml/issues/544
Make it more clear how to deserialize already parsed `toml_edit` entities with the new API.

### DIFF
--- a/crates/toml_edit/src/de/value.rs
+++ b/crates/toml_edit/src/de/value.rs
@@ -11,7 +11,6 @@ use crate::de::Error;
 /// # Example
 ///
 /// ```
-/// use serde::de::IntoDeserializer;
 /// use serde::Deserialize;
 ///
 /// #[derive(Deserialize)]
@@ -25,20 +24,11 @@ use crate::de::Error;
 ///     name: String,
 /// }
 ///
-/// let toml_str = r#"{ title = 'TOML Example', owner = { name = 'Lisa' } }"#;
-///
-/// let deserializer = toml_str
-///     .parse::<toml_edit::de::ValueDeserializer>()
-///     .unwrap();
+/// let value = r#"{ title = 'TOML Example', owner = { name = 'Lisa' } }"#;
+/// let deserializer = value.parse::<toml_edit::de::ValueDeserializer>().unwrap();
 /// let config = Config::deserialize(deserializer).unwrap();
 /// assert_eq!(config.title, "TOML Example");
 /// assert_eq!(config.owner.name, "Lisa");
-///
-/// let toml_item: toml_edit::Item = toml_str.parse().unwrap();
-/// let another_deserializer = toml_item.into_value().unwrap().into_deserializer();
-/// let another_config = Config::deserialize(another_deserializer).unwrap();
-/// assert_eq!(another_config.title, "TOML Example");
-/// assert_eq!(another_config.owner.name, "Lisa");
 /// ```
 pub struct ValueDeserializer {
     input: crate::Item,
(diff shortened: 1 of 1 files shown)


crates/toml_edit/src/de/value.rs
```diff
@@ -28,6 +28,9 @@ use crate::de::Error;
 /// assert_eq!(config.title, "TOML Example");
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
+///
+/// Already parsed [`crate::Value`] types can be turned into [`ValueDeserializer`]
+/// using [`serde::de::IntoDeserializer::into_deserializer`].
```
epage, line 33: nit: I feel like this should be before the `# Example` section.

At that point, it would probably be worthwhile to point out both relevant traits, something like
```md
To create a [`ValueDeserializer`], either use
- [`std::str::FromStr`]
- [`serde::de::IntoDeserializer::into_deserializer`] 
```

crates/toml_edit/src/de/value.rs
```diff
@@ -28,6 +28,9 @@ use crate::de::Error;
 /// assert_eq!(config.title, "TOML Example");
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
+///
+/// Already parsed [`crate::Value`] types can be turned into [`ValueDeserializer`]
```
epage, line 32: Does `crate::Value` get rendered literally in the rustdoc?  That  might not read all that well and we should probably take the `crate::` out of the link text

epage (commented):
Thanks!

epage:
> that naturally required the example adjustments to reflect both cases

Not necessarily.

crates/toml_edit/src/de/value.rs
```diff
@@ -21,12 +25,20 @@ use crate::de::Error;
...
+///     .unwrap();
+/// let config = Config::deserialize(deserializer).unwrap();
 /// assert_eq!(config.title, "TOML Example");
 /// assert_eq!(config.owner.name, "Lisa");
+///
+/// let toml_item: toml_edit::Item = toml_str.parse().unwrap();
+/// let another_deserializer = toml_item.into_value().unwrap().into_deserializer();
+/// let another_config = Config::deserialize(another_deserializer).unwrap();
+/// assert_eq!(another_config.title, "TOML Example");
+/// assert_eq!(another_config.owner.name, "Lisa");
```
epage, line 41: This should have a bit more narrative explaining whats going on, whether thats making them two separate examples with a bridge statement or a comment in them.

Or we can drop the second example

crates/toml_edit/src/de/value.rs
```diff
@@ -21,12 +25,20 @@ use crate::de::Error;
...
 ///
+/// let deserializer = toml_str
+///     .parse::<toml_edit::de::ValueDeserializer>()
+///     .unwrap();
+/// let config = Config::deserialize(deserializer).unwrap();
 /// assert_eq!(config.title, "TOML Example");
 /// assert_eq!(config.owner.name, "Lisa");
+///
+/// let toml_item: toml_edit::Item = toml_str.parse().unwrap();
+/// let another_deserializer = toml_item.into_value().unwrap().into_deserializer();
```
epage, line 38: Please parse directly into the `Value`.  For the role this plays, adding in `Item` confuses things.

SomeoneToIgnore:
In that case, let's remove the 2nd example indeed.
The conversion ways are enumerated better now, that's already good.

epage:
Thanks for doing this!

Issue #556, opened 2023-05-18
What are the semantics of the span in de::error?


konstin:
Would it be possible to add to the documentation what the semantics of the `Option<Range<usize>>` are that [`toml::de::Error::span()`](https://docs.rs/toml/latest/toml/de/struct.Error.html#method.span) returns? Is it a char offset or a byte offset, and under what conditions will it be None? I want to [integrate linting of toml files](https://github.com/charliermarsh/ruff/pull/4496) into our diagnostics system and it would be great to have clarity over how these indices behave we utf-8 content.

epage:
>  Is it a char offset or a byte offset

It is documented as [the index into the document](https://docs.rs/toml_edit/latest/toml_edit/de/struct.Error.html#method.span).  That would be a byte offset.

>  under what conditions will it be None

- If you deserialize from a `Document`, all span information is lost
- If you deserialize from a `&str`, then we should (nearly) alway have a span.  Any missing span is more likely to be a bug.

epage:
> Thanks! I unfortunately have encountered a lot of different things people mean by index, so this helps clarifying.

Within rust code for the `&str` type, [there is one canonical meaning](https://doc.rust-lang.org/std/primitive.str.html#impl-Index%3CI%3E-for-str)

PR #558, opened 2023-05-18
Extend de::Error::span() documentation

konstin:
Based on https://github.com/toml-rs/toml/issues/556#issuecomment-1553419512

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -65,7 +65,11 @@ impl Error {
         self.inner.message()
     }
 
-    /// The start/end index into the original document where the error occurred
+    /// The start/end index (byte offset) into the original document where the error occurred.
+    ///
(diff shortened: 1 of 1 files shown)


crates/toml/src/de.rs
```diff
@@ -65,7 +65,11 @@ impl Error {
         self.inner.message()
     }
 
-    /// The start/end index into the original document where the error occurred
+    /// The start/end index (byte offset) into the original document where the error occurred.
+    ///
+    /// If you deserialize from a `&str`, you should (nearly) always get a span. Any missing span is
+    /// more likely to be a bug. If you deserialize from a Document, all span information is lost
+    /// and this method will always return `None`.
```
epage, line 72: What I said to you was not how I would word things for a comment like this.  In general, I'm not sure I feel inclined to specify this at the moment

Issue #397, opened 2022-12-19
Support next TOML release


epage:
atm the next toml release contains
- Aligning inline tables with inline arrays
- Unicode bare identifiers
- etc

The unicode one will be a challenge as it might require us to stop using byte parsing and switch back to UTF-8 parsing which would slow us down a good chunk (#219).  Its not clear if they reused any existing unicode definitions like  [xid_continue](https://docs.rs/unicode-xid/latest/unicode_xid/), so I asked in toml-lang/toml#891

epage:
Note: this is blocked on the release being out but meant for our tracking of it

wolfv:
Has there been any progress on the inline-table with multiline support? I looked at some of the history but couldn't find anything yet. I think it's a nice usability improvement in certain cases.

We might be able to spend some time at @prefix-dev if the feature is desired.

epage:
At this time, I'd rather not be supporting unreleased TOML specs.

The issue for TOML 1.1 spec is toml-lang/toml#928

wolfv:
And if we would put it behind a feature flag like `toml-next` or `toml-1.1`?

epage:
That could get quite messy maintaining different subsets of the parsers like that.

We'd also need a good way of dividing the conformance tests to handle that.

I would also make it a `unstable-toml-1.1` flag because it could have breaking format changes.

Overall, this does not seem worth doing, especially because its not guaranteed that the approved parts of the format will stay as-is

rakuyoMo:
😮‍💨 The official Toml 1.1.0 version is progressing slowly, almost at a standstill. Many, many years have passed, and there's still no official release of 1.1.0. And some parsing libraries don't want to support features that haven't been released yet... It feels like I'm sinking into darkness, never to see the light again...

theoparis:
The golang toml parser already supports [toml 1.1 features like this](https://github.com/toml-lang/toml/issues/516) but I've been waiting for a rust equivalent for a few years now as well 😕 
My use case is a http proxy similar to Traefik which would need nested tables to simplify the toml configuration files.

epage:
There is talk of officially releasing the TOML 1.1 spec so I've posted #1079 for when it is released.

PR #564, opened 2023-05-30
Add support for additional escape characters

chrisduerr:
This patch adds support for the escapes `\xHH` and `\e` which will be part of the next TOML release.

---

I'm not sure what toml-rs' concrete policy on additions before the next official toml release is, but I found https://github.com/toml-rs/toml/issues/397 and thought contributions might be welcome.

These escapes would be extremely useful to me already, since they make specifying escape sequences like `\e[?1049h` significantly easier compared to using unicode `\u001b[?1029h`. So my goal is to make use of these upcoming TOML features even before official standardization.


epage:
As it isn't released yet, things can change.  I would like to wait until there is a release

Issue #504, opened 2023-01-31
single-line errors unexpectedly contain trailing newlines


davepacheco:
We have some tests in Dropshot to ensure that we produce useful messages when people provide invalid toml configuration files.  We noticed that with the upgrade from toml 0.5.11 to 0.7.0 (oxidecomputer/dropshot#580), simple single-line error messages wind up with trailing newlines.  Examples:

```
before:  `"missing field `level`"`
after:   `"missing field `level`\n"`
```

```
before:  `"unknown variant `everything`, expected one of `trace`, `debug`, `info`, `warn`, `error`, `critical`"`
after:   `"unknown variant `everything`, expected one of `trace`, `debug`, `info`, `warn`, `error`, `critical`\n"`
```

I think it's unexpected (and non-idiomatic?) to include trailing newlines on this sort of error message.

epage:
The problem is our errors may or may not be multi-line, depending on the context.   So the question is what would be idiomatic for multi-line errors.  I decided against making it conditional as that makes it hard to deal with in a general manner by callers.

rossmacarthur:
Also encountered this. Possibly it makes sense to use the alternate flag in `Display`?

E.g. I have seen some libraries use `format!("{:#}", err)` to display a pretty output, and `format!("{}", err)` to display a single line output. (or visa versa)

PR #568, opened 2023-06-02
Remove error trailing newline

zertosh:
Error `Display` impls should add their own trailing newline so the
caller can be in control about whether they print it or not. It is
common to print errors with `eprintln`. `toml`'s errors then end up
showing an extra newline.


crates/toml_edit/src/parser/errors.rs
```diff
@@ -131,9 +131,9 @@ impl Display for TomlError {
             }
             writeln!(f)?;
         }
-        writeln!(f, "{}", self.message)?;
+        write!(f, "{}", self.message)?;
         if !context && !self.keys.is_empty() {
-            writeln!(f, "in `{}`", self.keys.join("."))?;
+            write!(f, "\nin `{}`", self.keys.join("."))?;
```
zertosh, line 136: This is the substantive change. The rest are just tests

epage:
For a big change like this, let's discuss it in an issue first.  In particular, we have #504 for this

Issue #311, opened 2022-03-04
No `retain` for `Array` and `Table`/`TableLike`


jonhoo:
It's currently pretty painful to remove items matching a given condition from an array or table with `toml_edit` due to the lack of a `retain`-like method. You need one "pass" to gather the indices, and clone them if removing from a table, and then a second pass to remove each one, one by one. This is both unergonomic and inefficient. It'd be great to have `Array::retain` and `TableLike::retain` to make this kind of "match removal" easier!

epage:
Agreed though right now I won't be able to get to this as I'm focused on prep work for cargo-add going into cargo.  Contributions would be welcome!

marcospb19:
Love you both so I did #575.

`retain` should already be available to `Array` because it's a type alias to `Vec` :) .

marcospb19:
Unlinking #575 to this issue because I did it for the wrong crate :laughing:, I might try again for the right one :sweat_smile: .

marcospb19:
@epage you can close this.

I now realize that "Solves" [isn't a recognized keyword in GitHub](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for linking PRs to issues.

PR #576, opened 2023-06-28
feat(edit): add `retain` to `Array`, `Table` and `InlineTable`

marcospb19:
Solves #311.

### DIFF
--- a/crates/toml_edit/src/array.rs
+++ b/crates/toml_edit/src/array.rs
@@ -303,6 +303,20 @@ impl Array {
         }
     }
 
+    /// Retains only the values specified by the `keep` predicate.
+    ///
+    /// In other words, remove all values for which `keep(&value)` returns `false`.
+    ///
+    /// This method operates in place, visiting each element exactly once in the
+    /// original order, and preserves the order of the retained elements.
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&Value) -> bool,
+    {
+        self.values
+            .retain(|item| item.as_value().map(&mut keep).unwrap_or(false));
+    }
+
     fn value_op<T>(
         &mut self,
         v: Value,
--- a/crates/toml_edit/src/array_of_tables.rs
+++ b/crates/toml_edit/src/array_of_tables.rs
@@ -91,6 +91,20 @@ impl ArrayOfTables {
     pub fn remove(&mut self, index: usize) {
         self.values.remove(index);
     }
+
+    /// Retains only the elements specified by the `keep` predicate.
+    ///
+    /// In other words, remove all tables for which `keep(&table)` returns `false`.
+    ///
+    /// This method operates in place, visiting each element exactly once in the
+    /// original order, and preserves the order of the retained elements.
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&Table) -> bool,
+    {
+        self.values
+            .retain(|item| item.as_table().map(&mut keep).unwrap_or(false));
+    }
 }
 
 /// An iterator type over `ArrayOfTables`'s values.
--- a/crates/toml_edit/src/inline_table.rs
+++ b/crates/toml_edit/src/inline_table.rs
@@ -363,6 +363,24 @@ impl InlineTable {
             kv.value.into_value().ok().map(|value| (key, value))
         })
     }
+
+    /// Retains only the elements specified by the `keep` predicate.
+    ///
+    /// In other words, remove all pairs `(key, value)` for which
+    /// `keep(&key, &mut value)` returns `false`.
+    ///
+    /// The elements are visited in iteration order.
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&str, &mut Value) -> bool,
+    {
+        self.items.retain(|key, item| {
+            item.value
+                .as_value_mut()
+                .map(|value| keep(key, value))
+                .unwrap_or(false)
+        });
+    }
 }
 
 impl std::fmt::Display for InlineTable {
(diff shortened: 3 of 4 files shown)


marcospb19:
@epage I tried adding it to `TableLike`, but the generic for `FnMut(...) -> bool` made a compilation error, object safety requires that I don't put a generic in the `retain` method, I believe I could try with &dyn or &mut dyn, not sure if it works tho.

crates/toml_edit/src/array.rs
```diff
@@ -303,6 +303,19 @@ impl Array {
...
+    ///
+    /// In other words, remove all items for which `keep(&item)` returns `false`.
+    ///
+    /// This method operates in place, visiting each element exactly once in the
+    /// original order, and preserves the order of the retained elements.
+    pub fn retain<F>(&mut self, keep: F)
+    where
+        F: FnMut(&Item) -> bool,
+    {
+        self.values.retain(keep);
```
epage, line 316: This should accept `Value`, rather than `Item`.  The use of `Item` is an implementation detail and not exposed in the API more generally

crates/toml_edit/src/array.rs
```diff
@@ -303,6 +303,19 @@ impl Array {
         }
     }
 
+    /// Retains only the elements specified by the `keep` predicate.
+    ///
+    /// In other words, remove all items for which `keep(&item)` returns `false`.
+    ///
+    /// This method operates in place, visiting each element exactly once in the
+    /// original order, and preserves the order of the retained elements.
+    pub fn retain<F>(&mut self, keep: F)
```
epage, line 312: What about array of tables?

crates/toml_edit/src/inline_table.rs
```diff
@@ -363,6 +363,24 @@ impl InlineTable {
...
+
+    /// Retains only the elements specified by the `keep` predicate.
+    ///
+    /// In other words, remove all pairs `(internal_string, value)` for which
+    /// `keep(&internal_string, &mut value)` returns `false`.
+    ///
+    /// The elements are visited in iteration order.
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&InternalString, &mut Value) -> bool,
```
epage, line 375: I think we should use `&str` rather than `&InternalString`

marcospb19:
@epage I applied all changes and force-pushed accordingly.

Can I take the opportunity to ask two side-questions?
1. Do you prefer me to `mark as resolved`, or leave it to the reviewer to resolve?
2. Is force pushing in this context bad?

crates/toml_edit/src/array_of_tables.rs
```diff
@@ -91,6 +91,20 @@ impl ArrayOfTables {
...
+    ///
+    /// This method operates in place, visiting each element exactly once in the
+    /// original order, and preserves the order of the retained elements.
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&Value) -> bool,
+    {
+        self.values
+            .retain(|item| item.as_value().map(&mut keep).unwrap_or(false));
+    }
```
epage, line 107: This one should be using `as_table`

crates/toml_edit/src/table.rs
```diff
@@ -397,6 +397,24 @@ impl Table {
...
+    pub fn retain<F>(&mut self, mut keep: F)
+    where
+        F: FnMut(&str, &mut Value) -> bool,
+    {
+        self.items.retain(|key, item| {
+            item.value
+                .as_value_mut()
+                .map(|value| keep(key, value))
+                .unwrap_or(false)
+        });
```
epage, line 416: This one should be dealing with `Item`, not `Value`

epage:
> 1. Do you prefer me to `mark as resolved`, or leave it to the reviewer to resolve?

If its obvious, go ahead.  If its not, its nice to leave it open so we can both better track whether a change actually resolved it.

>     2. Is force pushing in this context bad?

I very much appreciate force-pushes

crates/toml_edit/src/array_of_tables.rs
marcospb19: Applied
```diff
-    /// In other words, remove all values for which `keep(&value)` returns `false`.
+    /// In other words, remove all tables for which `keep(&table)` returns `false`.
     ///
     /// This method operates in place, visiting each element exactly once in the
     /// original order, and preserves the order of the retained elements.
     pub fn retain<F>(&mut self, mut keep: F)
     where
-        F: FnMut(&Value) -> bool,
+        F: FnMut(&Table) -> bool,
     {
         self.values
-            .retain(|item| item.as_value().map(&mut keep).unwrap_or(false));
+            .retain(|item| item.as_table().map(&mut keep).unwrap_or(false));
```

crates/toml_edit/src/table.rs
marcospb19: Applied
```diff
-    /// In other words, remove all pairs `(key, value)` for which
-    /// `keep(&key, &mut value)` returns `false`.
+    /// In other words, remove all pairs `(key, item)` for which
+    /// `keep(&key, &mut item)` returns `false`.
     ///
     /// The elements are visited in iteration order.
     pub fn retain<F>(&mut self, mut keep: F)
     where
-        F: FnMut(&str, &mut Value) -> bool,
+        F: FnMut(&str, &mut Item) -> bool,
     {
-        self.items.retain(|key, item| {
-            item.value
-                .as_value_mut()
-                .map(|value| keep(key, value))
-                .unwrap_or(false)
-        });
+        self.items
+            .retain(|key, key_value| keep(key, &mut key_value.value));
```

This got a bit weird `.retain(|key, key_value|  ... key_value.value)`, because this Table maps `key -> key_value`.

epage:
Thanks!

PR #578, opened 2023-07-08
Add from_reader / from_slice / from_value function.

yassun7010:
Add `from_slice` and `from_reader` and ~`from_value`~ function  to be compatible with [serde_json](https://docs.rs/serde_json/latest/serde_json/index.html#functions),  [serde_yaml](https://docs.rs/serde_yaml/latest/serde_yaml/#functions).

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -133,6 +133,50 @@ where
     serde::Deserialize::deserialize(value)
 }
 
+/// Deserialize a slice of bytes into a type.
+///
+/// This function will attempt to interpret `v` as a TOML document and
+/// deserialize `T` from the document.
+///
+/// To deserializes TOML values, instead of documents, see [`ValueDeserializer`].
+///
+/// # Examples
+///
+/// ```
+/// use serde::Deserialize;
+///
+/// #[derive(Deserialize)]
+/// struct Config {
+///    title: String,
+///  owner: Owner,
+/// }
+///
+/// #[derive(Deserialize)]
+/// struct Owner {
+///     name: String,
+/// }
+///
+/// let config: Config = toml::from_slice(r#"
+///    title = 'TOML Example'
+///   [owner]
+///    name = 'Lisa'
+/// "#.as_bytes()).unwrap();
+///
+/// assert_eq!(config.title, "TOML Example");
+/// assert_eq!(config.owner.name, "Lisa");
+/// ```
+#[cfg(feature = "parse")]
+pub fn from_slice<'de, T>(v: &'de [u8]) -> Result<T, Error>
+where
+    T: serde::de::Deserialize<'de>,
+{
+    use serde::de::Error;
+
+    T::deserialize(Deserializer::new(
+        std::str::from_utf8(v).map_err(|err| Error::custom(err.to_string()))?,
+    ))
+}
+
 /// Errors that can occur when deserializing a type.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct Error {
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -92,6 +92,47 @@ where
     from_str(&buffer)
 }
 
+/// Deserialize a [`Value`] into a type.
+///
+/// This function will attempt to deserialize `T` from the [`Value`].
+///
+/// # Examples
+/// ```
+/// use serde::Deserialize;
+///
+/// #[derive(Deserialize)]
+/// struct Config {
+///    title: String,
+///   owner: Owner,
+/// }
+///
+/// #[derive(Deserialize)]
+/// struct Owner {
+///     name: String,
+/// }
+///
+/// let mut table = toml::value::Table::new();
+/// table.insert("title".to_owned(), toml::value::Value::String("TOML Example".to_owned()));
+///
+/// let mut owner = toml::value::Table::new();
+/// owner.insert("name".to_owned(), toml::value::Value::String("Lisa".to_owned()));
+/// table.insert("owner".to_owned(), toml::value::Value::Table(owner));
+///
+/// let value = toml::Value::Table(table);
+///
+/// let config: Config = toml::from_value(value).unwrap();
+///
+/// assert_eq!(config.title, "TOML Example");
+/// assert_eq!(config.owner.name, "Lisa");
+/// ```
+#[cfg(feature = "parse")]
+pub fn from_value<T>(value: crate::Value) -> Result<T, Error>
+where
+    T: serde::de::DeserializeOwned,
+{
+    serde::Deserialize::deserialize(value)
+}
+
 /// Errors that can occur when deserializing a type.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct Error {
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -45,6 +45,53 @@ where
     T::deserialize(Deserializer::new(s))
 }
 
+/// Deserializes a reader into a type.
+///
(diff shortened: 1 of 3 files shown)


crates/toml/src/de.rs
```diff
@@ -45,6 +45,138 @@ where
...
+///
+/// let value = toml::Value::Table(table);
+///
+/// let config: Config = toml::from_value(value).unwrap();
+///
+/// assert_eq!(config.title, "TOML Example");
+/// assert_eq!(config.owner.name, "Lisa");
+/// ```
+#[cfg(feature = "parse")]
+pub fn from_value<T>(value: crate::Value) -> Result<T, Error>
```
epage, line 129: Can we hold off on this one?  If interested in it, I'd recommend creating an issue for us to further discuss it.

Unlike json, TOML's top-level item must always be a table.  We previously had bugs and confusing behavior because we made `Value` the top-level within the API.  So I'm a little mixed on which it should be hear and would rather we just leave it off for now

crates/toml/src/lib.rs
```diff
@@ -165,7 +165,7 @@ mod table;
 
 #[cfg(feature = "parse")]
 #[doc(inline)]
-pub use crate::de::{from_str, Deserializer};
+pub use crate::de::{from_reader, from_slice, from_str, from_value, Deserializer};
```
epage, line 168: Is there a reason the `from`s are being added but not the `to`s?

See also #349

crates/toml/src/de.rs
```diff
@@ -45,6 +45,138 @@ where
...
+///     title = 'TOML Example'
+///     [owner]
+///     name = 'Lisa'
+/// "#.as_bytes()).unwrap();
+///
+/// assert_eq!(config.title, "TOML Example");
+/// assert_eq!(config.owner.name, "Lisa");
+/// ```
+#[cfg(feature = "parse")]
+pub fn from_reader<R, T>(rdr: R) -> Result<T, Error>
```
epage, line 80: We should probably have tests for these functions beyond the doctests

For example., `from_value` is using the table deserializer, rather than the value deserializer

epage:
I'm curious what your interest in these is.

Is there a case where this is more convenient?

My main concern is that someone will see these "lower level" functions and assume they are more performant when they aren't.

crates/toml/src/de.rs
yassun4dev: Yes, I remove `from_value` from this PR.

yassun4dev: This doctest does indeed use toml::Table instead of toml::Value.

However, I don't doubt the `from_value` implementation itself.

It is just a convenience function of `serde::Deserialize::deserialize(value)`.

Currently, I have no strong incentive to move forward with this, as I do not deal directly with toml::Value.

crates/toml/src/lib.rs
yassun4dev: Indeed!

I added to_writer/to_writer_pretty.

crates/toml/src/de.rs
yassun4dev: `from_value` has been put on hold.

Are there any items of concern in other functions?
I thought the test cases for the existing methods in toml were sufficient.

yassun4dev:
This is simply to provide compatibility with json and yaml implementations and reduce developer thinking time.

I'm making [conversion tools for data validation](https://github.com/yassun4dev/serde_valid), and I noticed that toml crate were not operation providing consistently.


I also aim to provide an easy way for developers to change data formats if they want to change it in their projects.

epage:
To me, that motivation feels very weak.  Yes, this provides consistency but we aren't being consistent in their implementation.

For `serde_json`, they either are independent implementations or the higher-level ones are implemented in terms of the lower level ones.

For `toml`, we are implementing the lower level ones in terms of the higher level.  As I said, I'm worried people will "drop down" to this API expecting better performance when it makes no difference.

yassun4dev:
Let me get my perception straight in order to decide if I should continue with this PR.

"`toml` is implemented the lower level ones in terms of the higher level" means that ser/de is using `toml_edit` internally in its implementation.
I recognize that the problem is that finer tuning is forgotten in the serialization or deserialization operations between toml strings and Rust structures.

When derializing, I believe that the string to structure can be uniquely determined.

On the other hand, when serializing, the output pattern cannot be uniquely determined and may not reflect user preferences
(e.g., how to represent arrays of tables).
And if user want to output customizing format, he needs to implemented `Serialize` for the Rust struct.

Is my rough understanding correct?
In the case of the above perception, I think users will implement their own `Serialize` according to their own convenience.

If the problem is not only with the format, you recognize the performance of the conversion process, I don't think many users understand the implementation of crates that well.
Performance tuning documentation will be needed.

epage:
Not quite.

I was referring to the fact that for `serde_json`, a `from_str` (higher level) is convenience wrapper around `from_slice` (lower level).  Contrast that with `toml` where `from_slice` (lower level) is a convenience wrapper around `from_str` (higher level).  People optimizing their code might assume working directly with a `File` / `from_reader` would be faster than a `File` and `from_slice` which would be faster than `File` and `from_str` when in fact the performance would all be the same (or worse).

yassun4dev:
When deserializing toml from a file, I think you end up needing to do the same thing I implemented with `from_reader`.

I suspect that users who use `from_reader` do not use it to improve performance, but simply to serialize from a file.

As for `from_slice`, it would certainly be better not to implement it, since toml requires utf-8 encoding in specification, so `from_str` is sufficient.

epage:
> When deserializing toml from a file, I think you end up needing to do the same thing I implemented with from_reader.

I think my own biases in how I write code are showing through here and I would like help in better understanding other view points, not just for this PR but for PRs generally.

When/why are you using `File` over alternatives?

For myself, I pretty much always use `std::fs::read_to_string` and `std::fs::write`.  I think last week is the first time in years that I used `File` and it was because I needed to append.

yassun4dev:
It seems natural for toml to handle `File` when reading from a configuration file.

Certainly relaying a `String` to a structure is the actual process (as is the implementation of `from_reader`), but I felt that going through a string was not the intent. As you say, this seems to be a bias.

nissaofthesea:
would it be possible to add `from_reader` without reading into a string first? imo the benefit of `from_reader` is to avoid the upfront cost of having the entire input in memory.

epage:
@yassun4dev 
> It seems natural for toml to handle `File` when reading from a configuration file.

Could you expand on this?  All of my own config handling in all of my applications call `std::fs::read` or `std::fs::read_to_string`.  In general, I find the `File` API a hassle to use and only do so when needed.  I'm not seeing why the use case of config files has an influence on what API is used.

@nissaofthesea 
> would it be possible to add `from_reader` without reading into a string first? imo the benefit of `from_reader` is to avoid the upfront cost of having the entire input in memory.

Doing so is out of scope at this time and I think you are highlighting why I'm hesitant to move forward with this.

yassun4dev:
@epage

I think I have come to understand why there is reluctance to introduce it.

The toml crate is focused on its function as a parse and isolates the io issue. I did not understand this when I raised the issue, but it is an excellent approach.

> > It seems natural for toml to handle File when reading from a configuration file.
>
> Could you expand on this? 

To explain a little, The File API is a hassle, but that process is hidden in the crate. I think this code expresses the purpose of mapping from toml format files to objects.

```rust
let data: MyStruct = serde_toml::from_reader(std::fs::File::open("sample.toml")?)?;
```

However, I would like to close this issue. My question has been resolved.

Thakns. I appreciate your explanation.

epage:
Any further discussion on this should be on #326 until this has been approved.

PR #610, opened 2023-09-13
feat(ser): serialize into any std::fmt::Write

tamird:
This allows more conveniently writing to files.

This change is backwards compatible (because `&mut String` is
`std::fmt::Write`) except that it changes the signatures of
`Serializer::new{,_pretty}` and `ValueSerializer::new`, which would
break users who pass these function as values.

Updates #326.


epage:
Please bring up your use case in the issue first and see how the discussion goes rather than potentially risking your time on wasted work or putting the extra social pressure of a PR on me of creating a PR.

I've been burned by this enough that I've taken a fairly strict approach to this and close PRs without much analysis because that should happen in the issue instead.

PR #618, opened 2023-10-03
fix typo: order *of* items

Miha-x64:


jsejcksn:
The [docs.rs](https://docs.rs/toml_edit/latest/toml_edit/) page still doesn't seem to include this fix.

epage:
Thanks for pointing that out.  Its been addressed in 052c0735dc99a8536b43ebad320c13765857821b

PR #620, opened 2023-10-10
Stop building unnecessary string

schubart:
Brought to you by the new Clippy lint `collection_is_never_read` which was added to Rust 1.70.

### DIFF
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -455,16 +455,13 @@ impl StringStyle {
 }
 
 fn infer_style(value: &str) -> (StringStyle, bool) {
-    // For doing pretty prints we store in a new String
-    // because there are too many cases where pretty cannot
-    // work. We need to determine:
+    // We need to determine:
     // - if we are a "multi-line" pretty (if there are \n)
     // - if ['''] appears if multi or ['] if single
     // - if there are any invalid control characters
     //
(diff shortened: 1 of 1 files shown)


crates/toml_edit/src/encode.rs
```diff
@@ -455,16 +455,11 @@ impl StringStyle {
...
-    // For doing pretty prints we store in a new String
-    // because there are too many cases where pretty cannot
-    // work. We need to determine:
+    // We need to determine:
     // - if we are a "multi-line" pretty (if there are \n)
     // - if ['''] appears if multi or ['] if single
     // - if there are any invalid control characters
-    //
-    // Doing it any other way would require multiple passes
-    // to determine if a pretty string works or not.
```
epage, line 466: Isn't this part of the comment still relevant?

schubart: I'm not sure. The comment seems to justify why a string was being built. It also explains what the function does, which is still true. I tried to get rid of the justification for the string and keep the rest.

Happy to keep those two lines if you think that's better.

epage: I think its referring to why the heuristics are so limited but not the biggest deal.

schubart: Thanks for explaining. I've brought the two lines back.

PR #632, opened 2023-10-24
Add instructions to set up local development environment

szabgab:
I think especially the part about pre-commit is important, but having detailed instructions about git can help more people.


epage:
Going from the base assumption that people don't know how to use git is not a good starting point for our contributing guide as that assumes too little knowledge.

As for pre-commit, that is an optional aid and not required.

PR #633, opened 2023-10-24
Add pre-commit hook to run fmt

szabgab:


epage:
I'm fairly ambivalent of this.  Most IDEs will have this specific check integrated and we have CI enforcing it.  I lean towards not adding this.

PR #651, opened 2023-11-13
Allow from_str to take Deserialize instead of DeserializeOwned

youknowone:
This change is introduced from here: https://github.com/toml-rs/toml/pull/457/files#diff-9dfce2824f02ba265b01f477a8d9821c7509063a170db31f465e3b176bbdab3dR40

Because DeserializeOwned allows fewer types than Deserialize, it introduces regression.
By looking [serde documentation](https://serde.rs/lifetimes.html#trait-bounds), `Deserialize` looking fits more for this function

### DIFF
--- a/crates/toml/src/de.rs
+++ b/crates/toml/src/de.rs
@@ -38,9 +38,9 @@
 /// assert_eq!(config.owner.name, "Lisa");
 /// ```
 #[cfg(feature = "parse")]
-pub fn from_str<T>(s: &'_ str) -> Result<T, Error>
+pub fn from_str<'a, T>(s: &'a str) -> Result<T, Error>
 where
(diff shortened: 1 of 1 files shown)


epage:
Quoting from #505

> From [#505 (comment)](https://github.com/toml-rs/toml/pull/505#discussion_r1092513171)
> 
> > Second, we are shifting a compile time error into a runtime error for the easy path while, as you pointed out, a workaround is available for those that want to borrow. In API design, we should be emphasizing the safe path with the easy path while ideally still supporting the unsafe route. The current solutions fits that to a T. I don't know why the serde docs make the recommendation they do but in this case, I see it being useful.
> > I'm going to go ahead and close this and ask any further discussion to happen in the appropriate issue (#490).

PR #654, opened 2023-11-24
Add Send + Sync for TableLike apis

AlseinX:
The current `TableLike` apis uses `dyn TableLike` without `Send` or `Sync`, while there is unlikely to be anything that blocks them from being `Send` or `Sync`.

To extend the usage of these features, I added `Send + Sync` on these apis.


epage:
Please create an issue for us to discuss this first.

epage:
Without an issue to disuss this and with this PR being fairly stale, I'm going to go ahead and close this.

Issue #368, opened 2022-11-09
Consider making the fields of `Spanned` public


Kixunil:
I think `Spanned` should be just a POD with all its fields public or at least the `value` field should be. Not being public makes matching extra annoying.

My use case: I have a toml that allows setting one of two fields only if another is either not set or set to `false` (they are contradictory) so I deserialize both fields and then I match on the tuple of them emitting an error if an invalid combination is present (and convert to a proper enum a valid one is). I'd love to make the error message look as nice as Rust error messages - having "foo specified here" and "bar specified here" messages pointing at the places where they are specified. I need span for this so I'm refactoring the match.

epage:
The only relevant issue I found in the old `toml-rs` repo was https://github.com/toml-rs/toml-rs/issues/436

I generally lean towards types being all `pub` or all private.  My main concern over exposing `Spanned`s fields is if we'd want to use an actual range type so we can clarify the bounds via the type system.

I wonder what the rest of the ecosystem does.
- `syn` is its own thing
- ariadne uses [range](https://docs.rs/ariadne/latest/ariadne/struct.Label.html)
- combine uses [start and end fields](https://docs.rs/combine/latest/combine/stream/span/struct.Span.html)

Kixunil:
To add to your list `codespan-reporting` uses range. Which I personally don't like much - semantically encoding "this is a span not some arbitrary range" is nice.

In my own crate I wrote `struct Span { start: usize, end: usize }` and conversions between `toml` and `codespan-reporting` types.

dtolnay:
I opened #664 to try to make progress by exposing a way to construct `Spanned`, even if the fields aren't public.

I would be happy for us to consider `Spanned::new(lo..hi, value)` or `Spanned::new(value).with_range(lo..hi)`, the latter not precluding construction with a better span type in the future if we settle on one.

dtolnay:
As I understand it, #635 adds a mechanism by which authors of a `Deserializer` impl can look at a particular `deserialize_struct` call and tell whether `Spanned<T>` is the thing being deserialized, and if so, feed it with the right range for `Spanned<T>` to get deserialized successfully.

That is not the missing piece in https://github.com/dtolnay/serde-untagged/issues/5. Serde-untagged implements a particularly clever `Visitor`, which the caller uses for implementing `Deserialize`. There is no <code>Deserialize<b>r</b></code> impl at play in that crate.

[...]

PR #664, opened 2023-12-19
Add Spanned::new(Range<usize>, T)

dtolnay:
I looked for why this might not exist and found #368, which is concerned with whether `Range<usize>` is really the right representation for the field type or if it should hold something else, like a custom `Span` type defined by serde_spanned.

The same concern might apply to a `new` constructor, and if so, an approach that doesn't preclude construction with a better type in the future might be preferred. For example `Spanned::new(value).with_range(range)`.

Without some API to construct a `Spanned` at a given range, I do not know of any way to deserialize an untagged enum containing `Spanned` fields. Context: https://github.com/dtolnay/serde-untagged/issues/5

### DIFF
--- a/crates/serde_spanned/Cargo.toml
+++ b/crates/serde_spanned/Cargo.toml
@@ -26,3 +26,9 @@ pre-release-replacements = [
 
 [dependencies]
 serde = { version = "1.0.145", optional = true }
+
+[dev-dependencies]
+serde = "1"
+serde_derive = "1"
+serde-untagged = "0.1"
+toml = { path = "../toml" }
--- a/crates/serde_spanned/src/spanned.rs
+++ b/crates/serde_spanned/src/spanned.rs
@@ -35,6 +35,63 @@ pub struct Spanned<T> {
 }
 
 impl<T> Spanned<T> {
+    /// Create a spanned value encompassing the given byte range.
+    ///
+    /// # Example
+    ///
+    /// Transposing a `Spanned<Enum<T>>` into `Enum<Spanned<T>>`:
+    ///
+    /// ```
+    /// use serde::de::{Deserialize, Deserializer};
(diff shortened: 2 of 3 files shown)


crates/serde_spanned/src/spanned.rs
```diff
@@ -35,6 +35,63 @@ pub struct Spanned<T> {
...
+    ///         Ok(match spanned.into_inner() {
+    ///             DependencyKind::Simple(simple) => Dependency::Simple(Spanned::new(range, simple)),
+    ///             DependencyKind::Detailed(detailed) => Dependency::Detailed(Spanned::new(range, detailed)),
+    ///         })
+    ///     }
+    /// }
+    /// #
+    /// # type DetailedDependency = std::collections::BTreeMap<String, String>;
+    /// ```
+    pub fn new(range: std::ops::Range<usize>, value: T) -> Self {
```
epage, line 91: > The same concern might apply to a new constructor, and if so, an approach that doesn't preclude construction with a better type in the future might be preferred. For example Spanned::new(value).with_range(range).

If we define a custom type, it can support `From<Range<usize>>` and this signature could be changed to `impl Into<Span>`, so I don't think we'd have compatibility concerns.

crates/serde_spanned/Cargo.toml
```diff
@@ -26,3 +26,9 @@ pre-release-replacements = [
 
 [dependencies]
 serde = { version = "1.0.145", optional = true }
+
+[dev-dependencies]
+serde = "1"
+serde_derive = "1"
+serde-untagged = "0.1"
+toml = { version = "0.8", path = "../toml" }
```
epage, line 34: This introduces a circular dependency which will prevent publishing.

If you drop `version`, everything should work

dtolnay: Good call! Done.

PR #668, opened 2024-01-28
Correct Table documentation

JustusAdam:
A tiny PR because I was just browsing the docs and noticed that this mentions the wrong implementation

WRT contributing guidelines: I'm happy to change the commits, although perhaps given the size of this fix, maybe we can just squash merge it.

### DIFF
--- a/crates/toml/src/table.rs
+++ b/crates/toml/src/table.rs
@@ -7,8 +7,11 @@ use crate::map::Map;
 use crate::Value;
 
 /// Type representing a TOML table, payload of the `Value::Table` variant.
-/// By default it is backed by a BTreeMap, enable the `preserve_order` feature
-/// to use a LinkedHashMap instead.
+///
+/// By default it entries are stored in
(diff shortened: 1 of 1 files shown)


crates/toml/src/table.rs
```diff
@@ -8,7 +8,7 @@ use crate::Value;
 
 /// Type representing a TOML table, payload of the `Value::Table` variant.
 /// By default it is backed by a BTreeMap, enable the `preserve_order` feature
-/// to use a LinkedHashMap instead.
+/// to use an IndexMap instead.
```
epage, line 11: Instead of specifying the data structures used, could you instead update it to specify the behavior (sorted by key, respect insertion-order)

JustusAdam: Should I also remove the reference to `BTreeMap` then and describe both just in terms of ordering (for consistency)?

epage: Yes, that'd be great!

JustusAdam:
Let me know what you think about this.

epage:
Thanks!

PR #706, opened 2024-03-22
docs: add example how to reject invalid keys

cgzones:
When parsing configuration files a common behavior is to reject invalid keys.


epage:
Thank you for wanting to make things better and improving what I assume is a problem you ran into.  However, this change is more about documenting `serde` than `toml` and I'd rather not get in the business of taking responsibility for teaching `serde` in this package nor do I think this scales for every `serde` related package to do so.

PR #722, opened 2024-04-30
docs: Remove confusing example version

tversteeg:
The version in the TOML example in the main documentation on docs.rs can be easily confused with the latest version of the crate, which is multiple versions newer.


epage:
I took your change and removed some more content since people might assume they should reach out to Alex based on that as well.

PR #726, opened 2024-05-07
docs: fix typo

mistydemeo:
Just a minor typo I noticed in the docs.


epage:
Its not a typo but an [abbreviation](https://en.wikipedia.org/wiki/If_and_only_if) but the difference is is small and causing confusion that we might as well change it.

mistydemeo:
Ahh, sorry!

PR #751, opened 2024-07-12
refactor: make repr more public

ip1981:
We are implementing a library to access TOML data via a "TOML path" like `foo[1].bar`, similar to [toml-cli](https://github.com/gnprice/toml-cli). Reusing the key parser provided by the public interface is tricky but possible and would be desired. But for additional feature requirements (like wildcards) we cannot use existing parser, and we cannot create a key in our parser because we need to set `Repr`.

This patch makes more public relevant parts for working with `Repr` so that we can create keys within our parser.

### DIFF
--- a/crates/toml_edit/src/key.rs
+++ b/crates/toml_edit/src/key.rs
@@ -53,7 +53,10 @@ impl Key {
         Self::try_parse_path(repr)
     }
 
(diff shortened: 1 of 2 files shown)


epage:
This is a significant API change in terms of exposing internals and should be discussed in an issue first.

PR #771, opened 2024-07-29
feat(datetime): impl Serialize/Deserialize for Date/Time

### LINKED ISSUE #763: Implement `Serialize` and `Deserialize` for `Date` and `Time`
The [`v1.0.0` TOML spec][toml-spec] allows omission of the time or date portion of a [local date-time][toml-spec-local-date-time] to represent a [local date][toml-spec-local-date] or a [local time][toml-spec-local-time] respectively. However, there is currently no implementation for `Serialize` or `Deserialize` for [`Date`][toml-rs-date] or [`Time`][toml-rs-time] in the `toml` crate. While it is possible for consumers of `toml` to use [`Datetime`][toml-rs-datetime], it would be better to be able to use `Date`/`Time` directly to more clearly communicate intent, prevent mistakes in making runtime assertions, and avoid the need to manually unwrap the `date`/`time` `Option`s.

**NOTE:** I am not suggesting we interpret a local time as a duration as suggested in #696. [RFC 3339][rfc-3339-1] explicitly does not cover time intervals.

[rfc-3339-1]: https://datatracker.ietf.org/doc/html/rfc3339#section-1
[toml-spec]: https://toml.io/en/v1.0.0
[toml-spec-local-date]: https://toml.io/en/v1.0.0#local-date
[toml-spec-local-time]: https://toml.io/en/v1.0.0#local-time
[toml-spec-local-date-time]: https://toml.io/en/v1.0.0#local-date-time
[toml-rs-date]: https://docs.rs/toml/0.8.16/toml/value/struct.Date.html
[toml-rs-time]: https://docs.rs/toml/0.8.16/toml/value/struct.Time.html
[toml-rs-datetime]: https://docs.rs/toml/0.8.16/toml/value/struct.Datetime.html

loqusion:
Closes #763

### DIFF
--- a/crates/toml_edit/src/parser/array.rs
+++ b/crates/toml_edit/src/parser/array.rs
@@ -1,12 +1,13 @@
 use winnow::combinator::cut_err;
 use winnow::combinator::delimited;
 use winnow::combinator::opt;
+use winnow::combinator::peek;
 use winnow::combinator::separated;
 use winnow::combinator::trace;
 
 use crate::parser::trivia::ws_comment_newline;
 use crate::parser::value::value;
-use crate::{Array, Item, RawString, Value};
+use crate::{Array, Item, RawString};
 
 use crate::parser::prelude::*;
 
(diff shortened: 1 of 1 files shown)


crates/toml/tests/testsuite/serde.rs
```diff
@@ -1270,6 +1270,114 @@ fn datetime_offset_issue_496() {
...
+
+    let document: Document = toml::from_str("date = 2024-01-01").unwrap();
+    assert_eq!(
+        document.date,
+        Date {
+            year: 2024,
+            month: 1,
+            day: 1
+        }
+    );
```
epage, line 1345: Why are we parsing and then deserializing, rather than directly creating a `Document`?

epage: nm, I missed something about the test

crates/toml/tests/testsuite/serde.rs
```diff
@@ -1270,6 +1270,114 @@ fn datetime_offset_issue_496() {
...
+            minute: 0,
+            second: 0,
+            nanosecond: 0,
+        }
+    );
+
+    let err = toml::from_str::<Document>("time = 2024-01-01T05:00:00").unwrap_err();
+    assert_data_eq!(
+        err.message(),
+        str!["expected only time portion of datetime"]
```
epage, line 1377: All commits should be passing.  When adding tests in a commit before doing feature work, make them show the current behavior. 

See https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#process

> Add tests in a commit before their feature or fix, showing the current behavior. The diff for the feature/fix commit will then show how the behavior changed, making it clearer to reviewrs and the community and showing people that the test is verifying the expected state.
>
>    e.g. https://github.com/clap-rs/clap/pull/5520

crates/toml_datetime/src/datetime.rs
```diff
@@ -513,6 +533,40 @@ impl<'de> de::Deserialize<'de> for Datetime {
...
+    where
+        D: de::Deserializer<'de>,
+    {
+        match Datetime::deserialize(deserializer)? {
+            Datetime {
+                date: Some(date),
+                time: None,
+                offset: None,
+            } => Ok(date),
+            _ => Err(de::Error::custom("expected only date portion of datetime")),
```
epage, line 548: I feel like [`Error::invalid_type`](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.invalid_type) would be more appropriate for these

epage:
Weird, unsure why thnis was closed

Issue #800, opened 2024-10-15
`de::Error` does not implement `source` method


dvzrv:
Hi! :wave: 

I'm using this crate in one of my projects [via confy](https://github.com/rust-cli/confy/blob/a7864eed4775f24fffc5dc58eb0aafd09f9f5f0b/src/lib.rs#L114) and really like its simplicity :ok_hand: 

Unfortunately I have been hit by a limitation. The `de::Error` type does not implement the `source` method, which prevents me from accessing its `inner`:

https://github.com/toml-rs/toml/blob/b05e8c489be8ebfc0acacc1ec3556d95cd8d2198/crates/toml/src/de.rs#L90

Is there a specific reason why this method is not implemented, or could this be changed? If so, I can provide a pull request if you want.

epage:
Why do you need access to `inner`?

`source` serves as a way to provide the error that caused the current error.  UIs generally will print both error messages.  In our case, `de::Error` is a newtype and reporting it as the `source` would be both incorrect and cause it to be listed twice in some UIs

dvzrv:
> Why do you need access to `inner`?
>
> `source` serves as a way to provide the error that caused the current error. UIs generally will print both error messages. In our case, `de::Error` is a newtype and reporting it as the `source` would be both incorrect and cause it to be listed twice in some UIs

I wanted to not print the outer Errors, but be able to print the specific TomlError.
Printing the chain was quite confusing in the use-case I had.

However, I think I bamboozled myself with a debug println (`println!("{:?}", error)`) into believing that I can't get to the message of the inner-most error (`toml::de::Error`, `toml_edit::de::Error` and finally `toml_edit::TomlError`).
Doing a `println!("{}", error)` on the `toml::de::Error` (wrapped by `ConfyError`) yields the desired behavior though, so I guess there is no need for any change after all.

Closing,  I guess implementations should rely on the fact that the `Display` implementation is delegated to the inner-most Error (in this case `toml_edit::TomlError`)?

epage:
> I wanted to not print the outer Errors, but be able to print the specific TomlError.
> Printing the chain was quite confusing in the use-case I had.
>
> ...
>
> Closing, I guess implementations should rely on the fact that the Display implementation is delegated to the inner-most Error (in this case toml_edit::TomlError)?

So to verify, are you saying that everything is fine and this can be closed?

wiktor-k:
[...]

PR #801, opened 2024-10-15
feat: Implement source for `toml::de::Error` and `toml_edit::de::Error`

dvzrv:
Both `toml::de::Error` and `toml_edit::de::Error` provide `inner`, but without implementing the `source` method of `std::error::Error` it is not possible to access it.

Fixes: https://github.com/toml-rs/toml/issues/800


epage:
We ask for Issues first *before* moving on to PRs.  Closing until we have settled on things in the issue.

PR #165, opened 2021-09-03
WIP: fix: Preserve dotted-key ordering

epage:
Fixes #163

PR #808, opened 2024-11-10
fix: Preserve dotted-key ordering

### LINKED ISSUE #163: Dotted key ordering isn't always preserved
```toml
hello.world = "a"
goodbye = "b"
hello.moon = "c"
```
gets turned into
```toml
hello.world = "a"
hello.moon = "c"
goodbye = "b"
```

mxndtaylor:
Fixes #163

Based on @epage's WIP work from #165.

Admittedly not the most efficient solution as it adds another `Vec::sort_by` to the `Table::append_values` and `InlineTable::append_values` functions, sorting the `Vec`s after the values have just been inserted into them.

The slightly more efficient `Vec::sort_by_key` is not necessarily possible since it might move non-dotted keys all together, thus again breaking the original document's order.

A better solution might be to insert the values into the values `Vec`s in the intended order, but this would be much more involved.

Note:
This change failed some tests from the `edit` functionality, specifically, `Table::{sort_values, sort_values_by}` and the `InlineTable` versions. So, I created a targeted fix for that, where it just sets the `Key.position` property to be the index in the iterator over its keys.

The implication here is that any future changes that involve key re-orderings on Tables will have to modify `Key.position` value, instead of just their order in the `items: IndexMap`. 

In fact, because of this, we could remove the ordering of the `IndexMap` completely now, and instead rely on the position value, thereby reducing the chances of future confusion.

Without a more comprehensive change like that, I have a feeling there might be an untested use case here somewhere, if something else I'm not aware of is modifying the `Key` order in `Table.items` or `InlineTable.items`, then this will now fail.

### DIFF
--- a/crates/toml_edit/src/inline_table.rs
+++ b/crates/toml_edit/src/inline_table.rs
@@ -2,7 +2,7 @@ use std::iter::FromIterator;
 
 use crate::key::Key;
 use crate::repr::Decor;
-use crate::table::{Iter, IterMut, KeyValuePairs, TableLike};
+use crate::table::{Iter, IterMut, KeyValuePairs, TableLike, sort_values_by_position};
 use crate::{InternalString, Item, KeyMut, RawString, Table, Value};
 
 /// Type representing a TOML inline table,
@@ -75,6 +75,7 @@ impl InlineTable {
                 _ => {}
             }
         }
+        sort_values_by_position(values);
     }
 
     /// Auto formats the table.
--- a/crates/toml_edit/src/inline_table.rs
+++ b/crates/toml_edit/src/inline_table.rs
@@ -2,7 +2,7 @@ use std::iter::FromIterator;
 
 use crate::key::Key;
 use crate::repr::Decor;
-use crate::table::{Iter, IterMut, KeyValuePairs, TableLike, sort_values_by_position};
+use crate::table::{Iter, IterMut, KeyValuePairs, TableLike, sort_values_by_position, modify_key_position};
 use crate::{InternalString, Item, KeyMut, RawString, Table, Value};
 
 /// Type representing a TOML inline table,
@@ -87,14 +87,14 @@ impl InlineTable {
     pub fn sort_values(&mut self) {
         // Assuming standard tables have their position set and this won't negatively impact them
         self.items.sort_keys();
-        for value in self.items.values_mut() {
+        modify_key_position(&mut self.items,|value| {
             match value {
                 Item::Value(Value::InlineTable(table)) if table.is_dotted() => {
                     table.sort_values();
                 }
                 _ => {}
             }
-        }
+        });
     }
 
     /// Sort Key/Value Pairs of the table using the using the comparison function `compare`.
@@ -123,14 +123,14 @@ impl InlineTable {
             };
 
         self.items.sort_by(modified_cmp);
-        for value in self.items.values_mut() {
+        modify_key_position(&mut self.items,|value| {
             match value {
                 Item::Value(Value::InlineTable(table)) if table.is_dotted() => {
                     table.sort_values_by_internal(compare);
                 }
                 _ => {}
             }
-        }
+        });
     }
 
     /// If a table has no key/value pairs and implicit, it will not be displayed.
--- a/crates/toml_edit/src/parser/inline_table.rs
+++ b/crates/toml_edit/src/parser/inline_table.rs
@@ -40,7 +40,8 @@ fn table_from_pairs(
     // Assuming almost all pairs will be directly in `root`
     root.items.reserve(v.len());
 
-    for (path, (key, value)) in v {
+    for (position, (path, (mut key, value))) in v.into_iter().enumerate() {
+        key.set_position(Some(position));
         let table = descend_path(&mut root, &path)?;
 
         // "Likewise, using dotted keys to redefine tables already defined in [table] form is not allowed"
@@ -162,6 +163,7 @@ mod test {
             r#"{a = 1e165}"#,
             r#"{ hello = "world", a = 1}"#,
             r#"{ hello.world = "a" }"#,
+            r#"{ hello.world = "a", goodbye = "b", hello.moon = "c" }"#,
         ];
         for input in inputs {
             dbg!(input);
--- a/crates/toml_edit/src/parser/mod.rs
+++ b/crates/toml_edit/src/parser/mod.rs
@@ -217,6 +217,10 @@ key = "value"
 "#,
             r#"hello.world = "a"
 "#,
+            r#"hello.world = "a"
+goodbye = "b"
+hello.moon = "c"
+"#,
             r#"foo = 1979-05-27 # Comment
 "#,
         ];
--- a/crates/toml_edit/src/key.rs
+++ b/crates/toml_edit/src/key.rs
@@ -32,6 +32,7 @@ pub struct Key {
     pub(crate) repr: Option<Repr>,
     pub(crate) leaf_decor: Decor,
     pub(crate) dotted_decor: Decor,
+    pub(crate) position: Option<usize>,
 }
 
 impl Key {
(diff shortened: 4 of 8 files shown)


mxndtaylor:
Just to add a note about the additional changes, I've switched the `Vec::sort_by` to `Vec::sort_by_key` using `usize::MAX` for the key when `position=None`.

Originally, I used `Vec::sort_by` because I thought that any constant value for `position=None` would cause problems as it wouldn't consider the insertion order, and just move those `Key`s to the end of the output. And this is largely true, however:
1.  every item parsed by this library will have `position=Some`
2.  all items manually constructed with this library will have `position=None` by default 
3. returning `Ordering::Equal` from `Vec::sort_by` for comparing `position=Some` to `position=None` does not result in a preserving of the order (nor does returning any of the other `Ordering` values)

If the `Key` objects are either all from the parser or all manually constructed, then the original `Vec::sort_by_key` and the previous `Vec::sort_by` implementations result in identical output.

If the `Key` objects are mixed, where some are from the parser and some are manually constructed, we would get behavior like this:

```rust
let doc = r#"foo.bar = 1

foo.baz.qwer = "a"
goodbye = { forever="no" }

foo.pub = 2
foo.baz.asdf = "b"
"#.parse<MutDocument>().unwrap();

let root = doc.as_table_mut();
root["foo"]["baz"]["zxcv"] = "CCC"
assert_data_eq!(doc.to_string(), r#"foo.bar = 1

foo.baz.qwer = "a"
foo.baz.asdf = "b"
foo.baz.zxcv = "CCC"
goodbye = { forever="no" }

foo.pub = 2
"#);
```

where `foo.baz.asdf` and `foo.baz.qwer` were moved when we made no modification to them. The fix is to consider `position=None` as simply "greater" than `position=Some` which can more easily be implemented with `Vec::sort_by_key`.


---
The only remaining problem is a somewhat odd one, that would not normally come up, which is to merge tables by iterating through the second table's keys manually like so:

```rust
let mut origin_doc = r#"
foo = 1
bar = 2
"#.parse::<DocumentMut>().unwrap();
let mut additional_doc = r#"
baz = 3
qux = 4
"#.parse::<DocumentMut>().unwrap();

let origin_table = origin_doc.as_table_mut();

// if we use origin_table.extend(addition_doc.iter()) this does not occur
// it only occurs when we directly get the key from its parsed 
additional_doc.iter().for_each(|(k, v)| {
    // we get the actually `Key` here with its position
    let (key, _) = additional_doc.get_key_value(k).unwrap();
    origin_table.insert_formatted(key, v.clone());
});

// baz (with position=Some(0) from the second table) gets inserted before 
// bar (with position=Some(1) from the first table)
assert_data_eq!(origin_doc.to_string(), r#"
foo = 1
baz = 3
bar = 2
qux = 4
"#);
```

But as noted, the more canonical process of using `Table.extend(Table.iter())` does not have this problem, so I think this is fine.

mxndtaylor:
not really sure if that comment should be a doc or not, maybe I should have just made it into a non-doc comment, but I just did what the linter said to instead, so hopefully that fixes it.

crates/toml_edit/src/inline_table.rs
epage, line 1: > In fact, because of this, we could remove the ordering of the IndexMap completely now, and instead rely on the position value, thereby reducing the chances of future confusion.

`Table` and `InlineTable` have the semantics of an `IndexMap`.  We need to preserve the order of content as it gets inserted. We could set a position on insertion.

Also, I'm unsure about allowing key positions to be moved between tables because the position is dependent on the other values within the same table and you can't easily translate that from one table to the next.

epage:
Could you clean up the commits for how you would want me to review and merge this?

My recommendation is to split up commits so that
- The first commit is a test (that **passes**), showing the current behavior
- Any code changes from the fix you can pull out into refactoring commits
- The actual fix which would also include updating the test to now pass

epage:
> If the Key objects are mixed, where some are from the parser and some are manually constructed, we would get behavior like this:

This is a very common case and is one the motivating reasons for this crate.  Grouping dotted keys, rather than treating a dotted key as something to add at the end, is a more natural behavior.  We currently get it for free.  We don't always guarantee a more natural way of doing things though.  However, if we were to change it, it would likely be a breaking behavior change.

mxndtaylor:
@epage I've reorganized the commits, and also added a fair bit of refactoring, to move a few duplicated code blocks into `internal_table.rs`. 

My intention with `internal_table` is that it is not part of the public api for the crate and that for any future changes that affect both the `Table` and `InlineTable` implementations of methods with the same name, that method's body could be moved to `internal_table.rs` and amended there.

Hopefully, that's matching with the appearances there.

> Grouping dotted keys, rather than treating a dotted key as something to add at the end, is a more natural behavior.

I see your point here, and I tried to keep the behavior of inserted dotted keys being grouped while only parsed dotted keys were kept where they were, but I could not find a way that I was happy with:

| Sorting Method | `position=None` behavior  | Result |
|--|--|--|
| `Vec::sort_by` | `Ordering::Equal` | _any_ key with `position=None` groups _all_ (or most in some cases, which is even stranger) dotted keys in that table |
| `Vec::sort_by` | `Ordering::Greater` or `Ordering::Less` | all keys with `position=None` go to top or bottom of the document |
| `Vec::sort_by_key` | constant `usize::MAX` | all keys with `position=None` go to the bottom of the document |

So, options 1 or 3 are probably the best in my eyes, so if 3 (what is currently implemented) is a breaking change I'll look into going with option 1 and see if I can narrow down some the weirdness, or find a novel approach.

crates/toml_edit/src/inline_table.rs
mxndtaylor: > We could set a position on insertion.

I think I missed this when I was initially reading your feedback, I did not investigate this as an option yet, I'll take a look at this in the next week or so when I can.

crates/toml_edit/src/parser/inline_table.rs
epage, line 1: nit: this commit type would be `test`

crates/toml_edit/src/parser/mod.rs
```diff
@@ -233,6 +233,55 @@ key = "value"
             assert_data_eq!(doc.to_string(), input.raw());
         }
     }
+    #[test]
```
epage: newline

crates/toml_edit/src/internal_table.rs
```diff
@@ -0,0 +1,179 @@
+use crate::{InlineTable, Item, Key, Table, Value};
+/// a place for helper methods supporting the table-like impls
+
+use crate::table::KeyValuePairs;
+
+/// GetTableValues provides the logic for displaying a table's items in their parsed order
+pub(crate) trait GetTableValues {
```
epage, line 7: It looks like this is a roundabout way to share code for manipulating `KeyValuePairs`.  Seems like the functionality should just be writtenm for `KeyValuePairs` and shared

imo if I were to do this, I would 
- Move `KeyValuePairs` into here
- Either rename the file to be about `KeyValuePairs` or rename the type (though not a fan of `InternalTable` though `InnerTable` might work)
- Either make `get_values` and `sort_values` either non-member functions that take `KeyValuePairs` or make `KeyValuePairs` a newtype (which might add its own complexity)

epage: Feel free to split this refactor out into its own commit.  It'll likely make it easier to review and get merged which can speed up reviewing both PRs.

Also, please consider if there are intermediate steps to the above that should be broken out into their own commit (e.g. adding a newtype of we go that route)

epage:
> So, options 1 or 3 are probably the best in my eyes, so if 3 (what is currently implemented) is a breaking change I'll look into going with option 1 and see if I can narrow down some the weirdness, or find a novel approach.

In case you were waiting on an answer, yes it is a breaking change.  I plan to make a breaking change "soon" but haven't had time to organize it and doubt you want to wait on it.

PR #823, opened 2025-01-10
Resolve Winnow deprecations

sam-peaches:
winnow-0.6.23 contains incompatible changes: "Located" no more top-crate re-export.

fixed by using "stream::Located" instead.


epage:
That was an accidental change and has been fixed in Winnow 0.6.24.

While we will eventually want to address the deprecation messages, this change is incomplete for that.

PR #825, opened 2025-01-18
docs(readme): Update example to prefer DocumentMut

slivingston:
Thanks for these great crates! I noticed that `Document` is deprecated but is used in [the example of the toml_edit/README](https://github.com/toml-rs/toml/blob/a07313eb8c1e9a08ecf46c108270940f12d4f0fc/crates/toml_edit/README.md). This pull request updates it to `DocumentMut`.

Note that [the similar example at the top of lib.rs](https://github.com/toml-rs/toml/blob/a07313eb8c1e9a08ecf46c108270940f12d4f0fc/crates/toml_edit/src/lib.rs#L9-L34) already uses `DocumentMut`, so <https://docs.rs/toml_edit/latest/toml_edit/#example> is good. However, the README appears in [the crates.io page](https://crates.io/crates/toml_edit). Thus this pull request is motivated.

### DIFF
--- a/crates/toml_edit/README.md
+++ b/crates/toml_edit/README.md
@@ -16,14 +16,14 @@ relative order* of items.
 ## Example
 
 ```rust
-use toml_edit::{Document, value};
+use toml_edit::{DocumentMut, value};
 
 fn main() {
     let toml = r#"
 "hello" = 'toml!' # comment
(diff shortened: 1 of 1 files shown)


epage:
Thanks!

PR #829, opened 2025-02-05
feat: impl Map get_key_value

tisonkun:
@epage 

This is brought from `serde_json`'s impl.

It helps when you want to tell the compiler that the key (&String) has the same lifetime to `&self`, while using the input `key: &Q` doesn't have this bound.

I found this while implementing https://github.com/cratesland/spath/pull/20/


tisonkun:
Also cc @dtolnay if you have spare time to double check that such cherry-pick is OK for toml's fork.

PR #839, opened 2025-03-05
docs: Distinguish Table/InlineTable

### LINKED ISSUE #838: Item::as_table is a beatrap
`Item::as_table` looks at first glance like the obvious method to call if you want to do a normal table value lookup.  But it returns `None` for `Item::Value(Value::InlineTable(..))`.

The call you should use instead for this is `as_table_like`.  That's a longer name (so less obvious).  There's nothing in the documentation of `as_table` to warn you about this.

I was tripped up by this because the author of a crate that now uses toml_edit [wrote the bug that this API induces](https://github.com/bkchr/proc-macro-crate/issues/57).  (I suspect from the history there that that project switched from another toml library, which presumably an `.as_table` function that gave `Some` for inline tables too.)

I suggest:

 1. Right away, improve the documentation for `.as_table` and `.is_table` (in `Item`)
 2. Next time we do an incompatible bump, rename `{as,is}_table` -> `{as,is}_outline_table` (or similar bikeshed) and `{as,is}_table_like` to `{as,is}_table`.

epage:
Fixes #838

### DIFF
--- a/crates/toml_edit/src/array.rs
+++ b/crates/toml_edit/src/array.rs
@@ -5,8 +5,7 @@ use crate::repr::Decor;
 use crate::value::{DEFAULT_LEADING_VALUE_DECOR, DEFAULT_VALUE_DECOR};
 use crate::{Item, RawString, Value};
 
-/// Type representing a TOML array,
-/// payload of the `Value::Array` variant's value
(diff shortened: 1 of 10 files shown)


ijackson:
Thanks

PR #846, opened 2025-03-22
feat(toml_edit): Implement `IntoDeserializer` for `Item`

sxyazi:
I'm working on a config parser based on `toml_edit` that requires the implementation of `IntoDeserializer` for interoperability with `serde`. 

I noticed that `toml_edit::Item` already has an `into_deserializer()` method, but it's not public. So, this PR exposes it by implementing the `IntoDeserializer` trait.

### DIFF
--- a/crates/toml_edit/src/de/item.rs
+++ b/crates/toml_edit/src/de/item.rs
@@ -32,7 +32,6 @@ use crate::de::Error;
 /// # }
 /// # }
 /// ```
-// TODO: fix the example
 pub struct ItemDeserializer {
     input: crate::Item,
     validate_struct_keys: bool,


epage:
After refreshing myself on this code and the motivations for its design, my concerns are
- I am uncertain what the right line is for what deserializers to provide.  Currently, we limit it to just the types that can be parsed (`Deserializer`, `ValueDeserializer`).  Do we support a `TableDeserializer`, `ArrayDeserializer`, or `KeyDeserializer`?
- We shouldn't be saying that a `ValueDeserializer` is an `Item`s deserializer.  This would require a refactor commit that pulls out an `ItemDeserializer` (private) that `ValueDeserializer` wraps with a follow up commit that publicly exposes `ItemDeserializer`.

Also, is there a reason you are using `toml_edit::Item` rather than `toml::Value`?

epage:
> In my case, I don't need them since I don't need to know the specific type of the content, and I just need to know it's an Item that can be deserialized into my struct T. Maybe for now, just keep them private as internal implementation details until someone needs them?

I'm trying to look for the principles of what the API should be, not handle one off cases.

sxyazi:
My knowledge of API design is very limited, so I might not be the right person to answer this. 

Before finding such principles, is there any chance of accepting this one-off PR? If not, I can close it - or I can create a separate issue for discussion if that would help. It's all up to you :)

sxyazi:
Hey @epage, I have separated `ValueDeserializer` into a new `ItemDeserializer`, and now `ValueDeserializer` is just a wrapper of `ItemDeserializer`. 

Let me know if there are any other changes needed.

crates/toml_edit/src/de/value.rs
```diff
@@ -35,20 +36,18 @@ use crate::de::Error;
 /// # }
 /// ```
 pub struct ValueDeserializer {
-    input: crate::Item,
-    validate_struct_keys: bool,
+    inner: ItemDeserializer,
 }
 
 impl ValueDeserializer {
     pub(crate) fn new(input: crate::Item) -> Self {
```
sxyazi, line 43: Perhaps it would be better to change `crate::Item` to `crate::Value` here, but I'm not sure if we should introduce more unrelated changes, as it might make the review more difficult. 

If you prefer, I can make this change in a new commit or a separate refactor PR.

PR #898, opened 2025-05-08
refactor(lexer): Pull out core logic from EOF

epage:


crates/toml_parse/src/lexer/mod.rs
```diff
@@ -40,14 +40,22 @@ impl<'i> Lexer<'i> {
...
         let mut vec = Vec::with_capacity(capacity);
-        vec.extend(self);
+
+        while let Some(byte) = self.stream.as_bytes().first() {
+            vec.push(process_token(*byte, &mut self.stream));
+        }
+
+        let eof_start = self.stream.current_token_start();
+        let eof_span = Span::new_unchecked(eof_start, eof_start);
+        vec.push(Token::new(TokenKind::Eof, eof_span));
```
epage, line 57: Turns out this seems to be slowing things down?

Issue #360, opened 2022-10-17
Add support for no_std environments


YtvwlD:
In the past I've written a bootloader which uses TOML as the format for the configuration file format.

The UEFI targets (`i686-unknown-uefi` and `x86_64-unknown-uefi`) provide `core` and `alloc` (among others), but they currently don't provide full `std` support.

https://github.com/toml-rs/toml-rs/pull/391 and https://github.com/toml-rs/toml-rs/pull/429 added support for those environments for the old `toml-rs` crate (thought they were never merged). I'd be nice to have support here.

YtvwlD:
So, do you have a better idea of the landscape now? :)

epage:
We likely will never support `core`.

Stepping through this...
- We would likely want to copy [what winnow does](https://github.com/winnow-rs/winnow/blob/main/src/lib.rs#L148) which simplifies some of the code.
- I think we could add the `std` feature without a breaking change as `std` wouldn't have anything exclusive to it and the types are the same  so someone should get the same behavior if they upgrade with `default-features = false`
- It might be some work making sure this is all propagated from our dependencies to `toml_edit` to `toml`.
- We might want to adopt `cargo hack` to help us in verifying this

My main remaining concern is any of the maintenance pain in dealing with this.  With `winnow`, which makes a lot more sense in core/alloc use cases, its been easy to mess up and use `std` types on accident and not find it for a bit (granted, I've not adopted `cargo hack` there)

Adrian8115:
I think this is really needed

luojia65:
We rely on `no_std` TOML parsing for RustSBI and other system software, so there's a strong need for a robust `no_std` toml solution.

epage:
`toml_parse` and `toml_write` are low level libraries that are or will be released that support `core`.

#952 changes `toml` to use `alloc`

epage:
(though this leaves out `toml_edit` as `toml` no longer depends on it)

PR #904, opened 2025-05-21
Make serde_spanned no_std compatible

benbrittain:


epage:
Could you expand on your use case for this without more `no_std` (#360)?  Looking at the reverse deps, I'm not seeing any other formats using this.

epage:
I got pinged on a jj discord thread about no_std jj.  I'm assuming that means no-std `toml` is needed and not just `serde_spanned.  That would mean this PR is insufficient.

I do not plan to make `toml_edit` `no_std` but I do hope to shift `toml` over to `toml_parse` and I'd allow `toml` to be `no_std` then.

benbrittain:
Sorry, about the initial lack of context. The ping on jj discord was intended as a low-effort FYI, not a rationale.

You can see a WIP change here where I started making `toml_edit` `no_std`
https://github.com/benbrittain/toml/commit/3d8b12bec3299cfaa9582d8b429a392c3a77730f as part of a greater `wasm32v1-none` experiment. It's not ready to land in it's entirely, but it also sounds like it's not something you'd like to upstream.

For what it's worth, I'd personally find having toml_edit be `no_std` + `alloc` very convenient, the only really ugly thing I've seen here is the lack of hashmaps in `alloc::collections`.

PR #924, opened 2025-06-05
chore: Prevent eol conversions

epage:


epage:
I think this is causing a test failure for #922.

PR #925, opened 2025-06-05
fix(datetime): Improve error message quality

epage:


crates/toml_datetime/src/datetime.rs
```diff
@@ -346,7 +346,7 @@ impl FromStr for Datetime {
             .map_err(|err| err.expected("year or hour"))?;
         let sep = lexer
             .next()
-            .ok_or(DatetimeParseError::new().expected("`-` or `:`"))?;
+            .ok_or(DatetimeParseError::new().expected("`-` (YYYY-MM) or `:` (HH:MM)"))?;
```
epage, line 349: While this may be confusing and we could do `hh:mm`, the general approach I'm taking is SCREAMING_CASE is for placeholders.

Issue #762, opened 2024-07-28
`toml_edit`: value is not written if table is constructed beforehand


tingerrr:
The following code simply discards the last write to the `tests` sub-key.

```rust
use toml_edit::{DocumentMut, value};

fn main() {
    let toml = "";
    let mut doc = toml.parse::<DocumentMut>().expect("invalid doc");
    // doc["tool"] = toml_edit::table();
    // doc["tool"]["typst-test"] = toml_edit::table();
    doc["tool"]["typst-test"]["tests"] = value("tests");
    println!("doc: \n{}\n", doc.to_string());
}
```

Produces the following output:
```toml
tool = { typst-test = { tests = "tests" } }
```

Uncommenting the second line produces:

```toml
tool = {}
```

Uncommenting the second and first line produces:
```toml
[tool]

[tool.typst-test]
tests = "tests"
```

When inspecting the table `tool.typst-test`, the value exists, it's simply not written.

epage:
I've updated the reproduction case to make it easier to follow
```rust
#!/usr/bin/env nargo
---
[dependencies]
toml_edit = "0.22"
---

use toml_edit::{DocumentMut, value};

fn main() {
    let toml = "";
    let mut doc = toml.parse::<DocumentMut>().expect("invalid doc");
    doc["tool"]["typst-test"]["tests"] = value("tests");
    println!("doc: \n{}\n", doc.to_string());

    let toml = "";
    let mut doc = toml.parse::<DocumentMut>().expect("invalid doc");
    doc["tool"] = toml_edit::table();
    doc["tool"]["typst-test"]["tests"] = value("tests");
    println!("doc: \n{}\n", doc.to_string());

    let toml = "";
    let mut doc = toml.parse::<DocumentMut>().expect("invalid doc");
    doc["tool"]["typst-test"] = toml_edit::table();
    doc["tool"]["typst-test"]["tests"] = value("tests");
    println!("doc: \n{}\n", doc.to_string());

    let toml = "";
    let mut doc = toml.parse::<DocumentMut>().expect("invalid doc");
    doc["tool"] = toml_edit::table();
    doc["tool"]["typst-test"] = toml_edit::table();
    doc["tool"]["typst-test"]["tests"] = value("tests");
    println!("doc: \n{}\n", doc.to_string());
}
```
```console
$ ./toml_edit-762.rs
warning: `package.edition` is unspecified, defaulting to `2021`
[...]

PR #932, opened 2025-06-06
fix(edit): Implicitly convert tables, rather than throwaway

epage:
I believe this will help with cases like #762.

I've had this branch sitting around for a while and I'm unsure why I never moved forward with it.  Maybe because this glosses over user bugs?  Except we allow this kind of sloppiness in other cases.


epage:
Really tempted to move from explicit table types to "prefer form" which this change would align with

PR #936, opened 2025-06-10
refactor: try literal strings before basic for TomlKey

ip1981:
Basic (`"in double quotes"`) strings are more powerful than literal (`'in single quotes'`) strings in representing keys, and unquoted (bare) strings are the least powerful. This patch makes sure we use the least possible representation for keys: try unquoted strings, then literal, then basic.


epage:
This is not a refactor but in fact a fairly impactful change.  This should be first discussed and agreed to in an Issue before moving on to a PR.

PR #945, opened 2025-06-12
feat(toml)!: Replace `toml_edit` with `toml_parse`

epage:
0_cargo::toml::document::2-features

| parser  | features               |  time |
|------------|-------------------------|---------|
| v0.5      |                                | 1032us |
| v0.5      | `preserve_order` | 924us |
| v0.8.23 |                                |  745us |
| `main`   |                                | 697us |
| `main`   | `preserve_order` | 589us |
| `de`       |                                | 702us |
| `de`       | `preserve_order` | 464us |
| `de`       | `unsafe`                | 706 |
| `de`       | `simd`                    | 702us |
| `de`       | all                           | 459us |

0_cargo::toml::manifest::2-features

| parser  | features               |  time |
|------------|-------------------------|---------|
| v0.5      |                                | 921us |
| v0.5      | `preserve_order` | 896us |
| v0.8.23 |                                |  644us |
| `main`   |                                | 590us |
| `main`   | `preserve_order` | 593us |
| `de`       |                                | 580us |
| `de`       | `preserve_order` | 445us |
| `de`       | `unsafe`                | 581us |
| `de`       | `simd`                    | 581us |
| `de`       | all                           | 439us |

1_map::toml::document::100

| parser  | features               |  time |
|------------|-------------------------|---------|
| v0.5      |                                | 99us |
| v0.5      | `preserve_order` | 100us |
| v0.8.23 |                                |  74us |
| `main`   |                                | 57us |
| `main`   | `preserve_order` | 58us |
| `de`       |                                | 38us |
| `de`       | `preserve_order` | 40us |
| `de`       | `unsafe`                | 37us |
| `de`       | `simd`                    | 39us |
| `de`       | all                           | 40us |

1_map::toml::document::100

| parser  | features               |  time |
|------------|-------------------------|---------|
| v0.5      |                                | 76us |
| v0.5      | `preserve_order` | 81us |
| v0.8.23 |                                |  66us |
| `main`   |                                | 45us |
| `main`   | `preserve_order` | 48us |
| `de`       |                                | 25us |
| `de`       | `preserve_order` | 34us |
| `de`       | `unsafe`                | 26us |
| `de`       | `simd`                    | 26us |
| `de`       | all                           | 34us |

PR #949, opened 2025-06-12
Add and document "perf" features

joshtriplett:
- **docs(edit): Document the "perf" feature**
- **feat(toml): Add "perf" feature, forwarding to `toml_parse?/simd`**

### DIFF
--- a/crates/toml/Cargo.toml
+++ b/crates/toml/Cargo.toml
@@ -33,6 +33,9 @@ std = ["indexmap?/std"]
 parse = ["dep:toml_parse", "dep:winnow"]
 display = ["std", "dep:toml_edit", "toml_edit?/display"]
 unsafe = ["toml_parse?/unsafe"]
+# Enable features that improve performance. For maximum performance, you may
+# also want to enable the "unsafe" feature.
+perf = ["toml_parse?/simd"]
 debug = ["std", "toml_parse?/debug", "dep:anstream", "dep:anstyle"]
 
 # Provide a method disable_recursion_limit to parse arbitrarily deep structures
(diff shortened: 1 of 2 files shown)


epage:
I'm currently going through breaking changes and as part of that, I'm tempted to remove the `perf` feature and directly expose `simd`.  Still trying to figure out what I want.  I am a bit looser with what I expose in `toml_edit` because it is less likely to show up in public APIs so I can be more free in breaking it.

joshtriplett:
@epage While you're making breaking changes, have you considered enabling simd by default, and letting people opt *out* of it if they really need to? I would expect far more people want the performance.

epage:
Looking at #945, I was wondering if `simd` made enough of a difference to be worth keeping (along with `unsafe`).

At least with `simd`, it only makes a performance difference and in a way that it should be safe to make it a default feature without "breaking" anyone.

I'm also split on how much defaults should be minimal vs all the bells and whistles.  Two very different users who can walk away with very different negative impressions based on the result.

joshtriplett:
> At least with simd, it only makes a performance difference and in a way that it should be safe to make it a default feature without "breaking" anyone.

That would be my reaction as well. I think there's a lot of value to defaulting to performance, rather than expecting people to opt in, when in practice the vast vast majority of people *should* have it enabled.

epage:
On a similar note, I've been wondering if the `display` feature should be off by default for `toml` as most people likely don't write out TOML.

joshtriplett:
> On a similar note, I've been wondering if the `display` feature should be off by default for `toml` as most people likely don't write out TOML.

On the theory that if you want to write out TOML you probably want `toml_edit`?

epage:
For json, a common use cases is to read and write as you are transmitting and receiving data.

For machines, TOML tends to be read only.  Yes, then there is also the number of ese cases left that need format preserving writes with `toml_edit`.  That leaves very few use cases for writing with `toml` (like `Cargo.lock`, or `cargo publish` writing a new `Cargo.toml`) and so I wonder if we should trim that out.

joshtriplett:
@epage Making the `display` feature no longer enabled by default, in `toml`, seems pretty reasonable by that argument. Along with a comment on that feature saying "if you're enabling this, you might want `toml_edit` instead".

PR #994, opened 2025-07-08
readme: mention toml_parser

tshepang:


epage:
This missed other crates and didn't have quite the right framing, so I pushed up a different change.

PR #996, opened 2025-07-09
feat(toml-edit): prettify DocumentMut in place

### LINKED ISSUE #999: Expose `Pretty` (or a helper) in toml-edit for formatting DocumentMuts
Hi, I’ve been using `toml-edit` in a workflow where I first convert a value into a `DocumentMut`, then apply some custom decorations and add a few comments before serializing the final TOML. However, if I just call `toml_edit::ser::to_document` and then `doc.to_string()`, the result is a bit… ugly.

Of course I could manually apply pretty decorations myself, but `ser::pretty::Pretty` already exists and does exactly that. My current workaround is:
1. call `to_string_pretty`
2. parse the resulting string back into a `DocumentMut`
3. apply my decorations
4. serialize again

Which feels redundant when all I really want is access to Pretty directly.

I propose moving `Pretty` into its own module alongside ser/de so it’s publicly (or not) accessible and add a `DocumentMut::fmt(&mut self)` or `pretty(&mut self)` helper method. The code is small, so extracting it should not impose a big difference in the lib size.

Related to #996.

saiintbrisson:
Hi, this patch closes #999 by moving `Pretty` to its own module alongside ser/de, which
shouldn't be a problem since it's rather small. The `DocumentMut::pretty` is a nice to
have but not needed, just having `Pretty` be accessible would be enough. Note that the
struct is kept private here.

I am unsure on where to add tests for this function, if any is needed.

EDIT:
Just saw
> Already have an idea? It might be good to first [create an issue](https://github.com/toml-rs/toml/issues/new) to propose it so we can make sure we are aligned and lower the risk of having to re-work some of it and the discouragement that goes along with that.

Sorry 😬, fortunately it took me 5min to do this so re-work is not that big of a problem 😅


crates/toml_edit/src/document.rs
```diff
@@ -190,6 +190,14 @@ impl DocumentMut {
     pub fn trailing(&self) -> &RawString {
         &self.trailing
     }
+
+    /// Apply a limited pretty formatting to this document.
+    ///
+    /// Use custom decors for advanced customization.
+    #[cfg(feature = "display")]
+    pub fn pretty(&mut self) {
```
saiintbrisson, line 198: Maybe call it `fmt` instead to play along the other functions in the code?

epage:
> Sorry 😬, fortunately it took me 5min to do this so re-work is not that big of a problem 😅

Except the question is "what is right for our API" and I would want that captured in an Issue, not a PR.  I'm marking this as a Draft until an Issue has been accepted.

saiintbrisson:
> > Sorry 😬, fortunately it took me 5min to do this so re-work is not that big of a problem 😅
> 
> Except the question is "what is right for our API" and I would want that captured in an Issue, not a PR. I'm marking this as a Draft until an Issue has been accepted.

Fair enough. created #999.

saiintbrisson:
Closes as per https://github.com/toml-rs/toml/issues/999#issuecomment-3062676362.

PR #1000, opened 2025-07-10
chore: Drop MSRV to 1.65

epage:


epage:
Our MSRV came from anstream which came from unicode-width

PR #1001, opened 2025-07-11
Avoid some infinite loops with 0.9

Keats:
I found this bug while trying to upgrade a tool to 0.9.
This is more a bug report with a hacky fix but I wanted to see what it would take the make the test pass. 
Feel free to only take the test, I didn't want to really dig into the lexer/parser itself

### DIFF
--- a/crates/toml/tests/fixtures/invalid/not-toml/deb-like.toml
+++ b/crates/toml/tests/fixtures/invalid/not-toml/deb-like.toml
@@ -1,0 +1,6 @@
+Package: R6
+Version: 2.5.1
+Depends: R (>= 3.0)
+Suggests: testthat, pryr
+NeedsCompilation: no
+License: MIT + file LICENSE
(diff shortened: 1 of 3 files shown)


crates/toml/tests/fixtures/invalid/not-toml/deb-like.toml
epage, line 1: These should go within the `ext` folder to distinguish them from the official test cases

epage, line 1: test cases in this folder need to be mirrored over into `toml_edit`

epage:
> I found this bug while trying to upgrade a tool to 0.9.
This is more a bug report with a hacky fix but I wanted to see what it would take the make the test pass.

If this is a bug report, please open an issue rather than a PR with a non-starter fix (hacky is putting this mildly)

Issue #812, opened 2024-11-15
Add support for outputting integers as hexadecimal


MatthewMorozov:
When using serde to serialize a data structure containing integers (u32, i16, etc) to TOML format, I would like the fields to be written in the generated TOML file in hexadecimal format. TOML format supports hexadecimal numbers, and this feature would make it easier to read fields such as 32-bit addresses. This may could be an option such as `#[serde_with_format = "{:0X}"]` above integer-type fields.

epage:
Some prior discussion at #781.

I don't see us supporting customization of formatting through serde.  Serde doesn't natively support it.  To support this, we'd need a newtype that serializes to a struct that `toml` recognizes as an integer with formatting.  We'd support deserializing from either the struct or a number.  This has some odd cases in serde, particularly that its not compatible with any other serialization format.  Not just that the formatting gets ignored but that the number won't be sererialized correctly.

We can support this in `toml_edit` and people could deserialize to a `DocumentMut` and then apply a custom formatter.  We'd just need to work out the API for customizing an integer's repr.

epage:
In #781, I mention the following API ideas
- Separate APIs for each.  When applying botth, we'd either need to store the format style (would rather not do this) or parse the repr to see what was done and re-generate
- A single format call for integers, taking in a non-exhaustive `IntegerFormat`
- A single format call for any scalar, taking in a non-exhaustive `ScalarFormat`.  This could then store all the settings for string formatting (see #590), etc

#837 was posted with `Formatted<i64>::new_hex_lower`, etc.  This has the benefit of communicating whether the formatting was applied.  It has the downsides of
- Requires creating a new formatted
- No `Into`s for `Formatted<i64>`
- No indication of how we would handle digit separator

Should we also allow reflecting on what the current style is?

epage:
> Should we allow customization of digit separators?
>
> - I don't know. I anticipate significant demand for this feature. However, I don't believe there exists a set of separator styles that covers the majority of use cases. Do users want 1234_5678, 12_34_56_78, or 12_345_678?

In your examples, they are always grouped by a standard convention.  For example, hex is usually grouped by 4s while decimal is usually grouped by 3s, starting from the right.

This could easily be handled by just having a "group by" option.  This can trivially be implemented by post-processing a string (starting from the back, insert a `_` after every N digits).

[...]

PR #1004, opened 2025-07-11
feat(writer): Implement TomlInteger

pastelmind:
Implements `TomlInteger` and `TomlIntegerFormat` following the discussion at #812. This supersedes #837.

### DIFF
--- a/crates/toml_writer/src/integer.rs
+++ b/crates/toml_writer/src/integer.rs
@@ -1,0 +1,194 @@
+use core::fmt::{self, Display};
+
+/// N must be an integer type.
+#[derive(Copy, Clone, Debug)]
+pub struct TomlInteger<N> {
+    value: N,
+    radix: Radix,
+}
+
+impl crate::WriteTomlValue for TomlInteger<u8> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+impl crate::WriteTomlValue for TomlInteger<i8> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+impl crate::WriteTomlValue for TomlInteger<u16> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+impl crate::WriteTomlValue for TomlInteger<i16> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+impl crate::WriteTomlValue for TomlInteger<u32> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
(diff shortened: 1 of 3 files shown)


crates/toml_writer/src/integer.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct TomlIntegerFormat {
+    radix: Radix,
+    _separators: (), // Placeholder for future use
+}
```
epage: As this is the primary type people will interact with, please put this at the top of the file

crates/toml_writer/src/integer.rs
```diff
@@ -0,0 +1,194 @@
...
+impl crate::WriteTomlValue for TomlInteger<isize> {
+    fn write_toml_value<W: crate::TomlWrite + ?Sized>(&self, writer: &mut W) -> fmt::Result {
+        write_toml_value(self.value, &self.radix, writer)
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct TomlIntegerFormat {
+    radix: Radix,
+    _separators: (), // Placeholder for future use
```
epage: Is there a reason for us to have this field?

crates/toml_writer/src/integer.rs
```diff
@@ -0,0 +1,194 @@
...
+                if value < 0 {
+                    return None;
+                }
+            }
+        }
+
+        Some(TomlInteger {
+            value,
+            radix: self.radix,
+        })
```
epage: Should we just copy `TomlIntegerFormat` into `TomlInteger`?

crates/toml_writer/src/integer.rs
pastelmind: If I remove the `separators` field, Clippy complains about the struct update syntax in `TomlIntegerFormat::as_decimal()`. I wanted to keep the struct update syntax to make the intent clear. Do you want me to remove it?

crates/toml_writer/src/integer.rs
```diff
@@ -0,0 +1,194 @@
+use core::fmt::{self, Display};
+
+/// N must be an integer type.
```
epage: The first line of a docstring is generally used as a summary.  We should probably use [Display](https://doc.rust-lang.org/std/path/struct.Display.html) for inspiration.

Also considering people can't create a `TomlInteger` directly, I don't think we need to document the constraints for `N`

pastelmind: I'll just remove the docstring for now.

crates/toml_writer/tests/integer.rs
epage, line 1: Please use `snapbox`s `assert_data_eq!` with the expected value being `str![]` for tests, like we do with `float.rs`

crates/toml_writer/src/integer.rs
epage: For my builders, I tend to just make things `mut self` and just assign in.  imo that is much clearer intent and we don't risk the compiler doing something weird (copying the whole thing when just a field needs to be modified)

crates/toml_writer/src/integer.rs
epage: I think it would be good to help people know more explicitly the relationship of the types than having to read their functions to figure it out

epage: ... says the person who didn't do that for the string stuff

pastelmind:
I'm not familiar with snapbox, so feel free to suggest a different test setup.

epage:
tests are failing when `--no-default-features` is applied

pastelmind:
Sorry, it should run fine now

epage:
For future reference, I prefer people to maintain a PR's commits for how they would like them to look when merged, rather than a commit per PR update.  There is value in having and merging multiple commits in a PR that can get lost with squashing.

PR #1014, opened 2025-07-21
fix(parser): Improve bad quote error messages

### LINKED ISSUE #1013: v0.9 parse errors for invalid string missing leading `"` regressed
See https://github.com/zanieb/toml-regression-2 for a simple reproduction

In 0.8.19

```
TOML parse error at line 3, column 16
  |
3 | tqdm = { url = invalid url to tqdm-4.66.0-py3-none-any.whl" }
  |                ^
invalid string
expected `"`, `'`
```

In 0.9.2

```
TOML parse error at line 3, column 28
  |
3 | tqdm = { url = invalid url to tqdm-4.66.0-py3-none-any.whl" }
  |                            ^
missing comma between key-value pairs, expected `,`
```

I think the former error message is more helpful for users, though I'm sure it's plausible you could contrive cases where it's wrong :)

Originally discovered at https://github.com/astral-sh/uv/pull/14571#discussion_r2213785912

epage:
Fixes #1013

### DIFF
--- a/crates/toml/tests/snapshots/invalid/array/missing-separator-02.stderr
+++ b/crates/toml/tests/snapshots/invalid/array/missing-separator-02.stderr
@@ -1,12 +1,5 @@
-TOML parse error at line 1, column 15
-  |
-1 | wrong = [ 1 2 3 ]
-  |               ^
-missing comma between array elements, expected `,`
-
----
 TOML parse error at line 1, column 11
   |
 1 | wrong = [ 1 2 3 ]
-  |           ^^^
+  |           ^^^^^
 string values must be quoted, expected literal string
--- a/crates/toml/tests/snapshots/invalid/inline-table/no-comma-02.stderr
+++ b/crates/toml/tests/snapshots/invalid/inline-table/no-comma-02.stderr
@@ -23,4 +23,4 @@ TOML parse error at line 1, column 26
   |
 1 | arrr = { comma-missing = true valid-toml = false }
   |                          ^^^^^^^^^^^^^^^
-invalid boolean, expected `true`
+string values must be quoted, expected literal string
--- a/crates/toml_parser/src/decoder/scalar.rs
+++ b/crates/toml_parser/src/decoder/scalar.rs
@@ -655,12 +655,16 @@ pub(crate) fn decode_symbol<'i>(
     error: &mut dyn ErrorSink,
 ) -> ScalarKind {
     if raw.as_str() != symbol {
-        error.report_error(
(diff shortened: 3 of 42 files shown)


crates/toml/tests/snapshots/invalid/array/missing-separator-02.stderr
```diff
@@ -1,12 +1,5 @@
...
-1 | wrong = [ 1 2 3 ]
-  |               ^
-missing comma between array elements, expected `,`
-
----
 TOML parse error at line 1, column 11
   |
 1 | wrong = [ 1 2 3 ]
-  |           ^^^
+  |           ^^^^^
```
epage, line 4: This extends the `[true false]` problems to `[1 2 3]`

crates/toml_parser/src/decoder/scalar.rs
```diff
@@ -360,6 +360,8 @@ pub(crate) fn decode_datetime_or_float_or_integer<'i>(
 
     if rest.starts_with("-") || rest.starts_with(":") {
         decode_as_is(raw, ScalarKind::DateTime, output, error)
+    } else if rest.contains(" ") {
+        decode_invalid(raw, output, error)
```
epage, line 364: We need to check the performance implications of this change

crates/toml/tests/snapshots/invalid/inline-table/no-comma-02.stderr
```diff
@@ -23,4 +23,4 @@ TOML parse error at line 1, column 26
   |
 1 | arrr = { comma-missing = true valid-toml = false }
   |                          ^^^^^^^^^^^^^^^
-invalid boolean, expected `true`
+string values must be quoted, expected literal string
```
epage, line 26: Would be good if we didn't merge the atoms in this case.

crates/toml/tests/snapshots/invalid/array/missing-separator-02.stderr
epage: Hmm, maybe it was already a problem, the test cases just cover different cases

epage: ie if you have one missing comma, it shows up weird.  If you have 2+ missing commas, then you get a hint as to the problem.  The latter case is what we changed.

crates/toml_parser/src/decoder/scalar.rs
epage: Looks like this is neutral within the cargo `2-features` benchmark

PR #1016, opened 2025-07-22
`ImDocument` hyperlink in `span()` should now point to `Document`

THEGOLDENPRO:
Hi, I glanced at the documentation and noticed that the `span()` function's docs were mentioning the now deprecated `ImDocument` instead of it's replacement struct  `Document`, so I went ahead and quickly corrected that.

Kind regards

### DIFF
--- a/crates/toml_edit/src/array.rs
+++ b/crates/toml_edit/src/array.rs
@@ -88,7 +88,7 @@ impl Array {
 
     /// The location within the original document
     ///
-    /// This generally requires an [`Document`][crate::Document].
+    /// This generally requires a [`Document`][crate::Document].
     pub fn span(&self) -> Option<std::ops::Range<usize>> {
         self.span.clone()
     }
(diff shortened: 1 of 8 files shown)


crates/toml_edit/src/array.rs
```diff
@@ -88,7 +88,7 @@ impl Array {
 
     /// The location within the original document
     ///
-    /// This generally requires an [`ImDocument`][crate::ImDocument].
+    /// This generally requires an [`Document`][crate::Document].
```
epage:
```suggestion
    /// This generally requires a [`Document`][crate::Document].
```
ditto for the others

THEGOLDENPRO: Alright done.

epage:
Note in the future that I'd prefer commits to be reviewed how they will be merged, meaning you are free and encouraged to edit them.

PR #1032, opened 2025-08-24
refactor: add docs and named fields to Data struct

KushalMeghani1644:


epage:
Could you help me understand the motivation for looking at this code and making this change/?

KushalMeghani1644:
The purpose of my PR is to mainly make subtle changes that improve the readability of the code with more clearer docs.

PR #1036, opened 2025-09-14
refactor: Switch serde dependency to serde_core

jplatte:
Improve build parallelism :)


crates/toml_datetime/src/ser.rs
```diff
@@ -50,14 +50,14 @@ impl DatetimeSerializer {
         Self { value: None }
     }
 
-    /// See [`serde::ser::SerializeStruct::serialize_field`]
+    /// See [`serde_core::ser::SerializeStruct::serialize_field`]
```
jplatte, line 53: Is this okay? Linking to `serde` instead would require that to be a dependency. I can add it as an optional behind a `__private_docs` feature or somesuch if you want (that's what we do in axum to link to things that aren't [direct] dependencies).

crates/toml_edit/src/de/value.rs
```diff
@@ -8,14 +8,15 @@ use crate::de::TableDeserializer;
 /// Deserialization implementation for TOML [values][crate::Value].
 ///
 /// Can be created either directly from TOML strings, using [`std::str::FromStr`],
-/// or from parsed [values][crate::Value] using [`serde::de::IntoDeserializer::into_deserializer`].
+/// or from parsed [values][crate::Value] using
+/// [`IntoDeserializer::into_deserializer`][serde_core::de::IntoDeserializer::into_deserializer].
```
jplatte, line 12: Same Q here. This one I decided to handle slightly differently (shortening the displayed path). No strong opinion from me though.

crates/toml_datetime/src/ser.rs
paolobarbolini: Wouldn't having `serde` as a dev dependency work?

jplatte: No, dev-dependencies are only loaded for building / running doctests, not for generating docs.

crates/toml/Cargo.toml
```diff
@@ -49,7 +49,7 @@ unbounded = []
 preserve_order = ["dep:indexmap", "std"]
 
 [dependencies]
-serde = { version = "1.0.145", default-features = false, features = ["alloc"], optional = true }
+serde_core = { version = "1.0.145", default-features = false, features = ["alloc"], optional = true }
```
epage:
```suggestion
serde_core = { version = "1.0.221", default-features = false, features = ["alloc"], optional = true }
```

crates/toml_datetime/Cargo.toml
```diff
@@ -25,12 +25,12 @@ pre-release-replacements = [
...
-alloc = ["serde?/alloc"]
-std = ["alloc", "serde?/std"]
-serde = ["dep:serde"]
+alloc = ["serde_core?/alloc"]
+std = ["alloc", "serde_core?/std"]
+serde = ["dep:serde_core"]
 
 [dependencies]
-serde = { version = "1.0.145", default-features = false, optional = true }
+serde_core = { version = "1.0.145", default-features = false, optional = true }
```
epage:
```suggestion
serde_core = { version = "1.0.221", default-features = false, optional = true }
```

crates/toml_datetime/src/ser.rs
epage: I'm good with it

crates/toml_edit/Cargo.toml
```diff
@@ -41,7 +41,7 @@ unbounded = []
 [dependencies]
 indexmap = { version = "2.3.0", features = ["std"] }
 winnow = { version = "0.7.10", optional = true }
-serde = { version = "1.0.145", optional = true }
+serde_core = { version = "1.0.145", optional = true }
```
epage:
```suggestion
serde_core = { version = "1.0.221", optional = true }
```

crates/toml_datetime/src/ser.rs
epage: And thanks for calling this out as its easy to miss something like this among the more monotonous details!

epage:
Please keep in mind that I prefer a clean commit history and am perfectly fine with editing of commits

epage:
Thanks!

PR #1051, opened 2025-10-10
Fix build status badges

atouchet:
I do not see any workflows named "Continuous integration" so I am assuming ci.yml is the correct file to use.


epage:
Thanks for reporting this!  I went ahead and removed those badges in 6df349bf to match my project template https://github.com/epage/_rust

PR #1058, opened 2025-10-31
perf(toml_edit): reduce llvm-lines in `Option::and_then` calls

dishmaker:
158 less llvm-lines.

- Might reduce compile time of `toml_edit`.

- Removing iterators might be more helpful, like in https://github.com/rust-lang/log/pull/709
  Especially I don't know where `Iterator::try_fold` is coming from.
  Big improvements were made by removing iterators in https://github.com/RustCrypto/formats/pull/1810

Before:
```shell
$ cargo llvm-lines | head -n 20
  Lines                Copies              Function name
  -----                ------              -------------
  58354                2328                (TOTAL)
   1996 (3.4%,  3.4%)    24 (1.0%,  1.0%)  core::iter::traits::iterator::Iterator::try_fold
   1276 (2.2%,  5.6%)     1 (0.0%,  1.1%)  alloc::str::join_generic_copy
   1060 (1.8%,  7.4%)     4 (0.2%,  1.2%)  core::slice::sort::stable::quicksort::stable_partition
    970 (1.7%,  9.1%)    43 (1.8%,  3.1%)  core::option::Option<T>::and_then
    688 (1.2%, 10.3%)    30 (1.3%,  4.4%)  core::option::Option<T>::map
    586 (1.0%, 11.3%)    24 (1.0%,  5.4%)  core::option::Option<T>::unwrap_or_else
    575 (1.0%, 12.3%)     5 (0.2%,  5.6%)  toml_edit::encode::encode_formatted
    565 (1.0%, 13.2%)     1 (0.0%,  5.7%)  core::num::<impl i64>::from_ascii_radix
    540 (0.9%, 14.1%)    15 (0.6%,  6.3%)  <I as core::iter::traits::iterator::Iterator::advance_by::SpecAdvanceBy>::spec_advance_by
    523 (0.9%, 15.0%)    15 (0.6%,  7.0%)  core::iter::traits::iterator::Iterator::nth
    488 (0.8%, 15.9%)     2 (0.1%,  7.0%)  core::slice::sort::shared::smallsort::small_sort_general_with_scratch
    488 (0.8%, 16.7%)     2 (0.1%,  7.1%)  core::slice::sort::stable::drift::sort
    464 (0.8%, 17.5%)     8 (0.3%,  7.5%)  core::option::Option<T>::get_or_insert_with
    457 (0.8%, 18.3%)    12 (0.5%,  8.0%)  <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
[...]

PR #1062, opened 2025-11-04
perf(toml_edit): Rewrite array.rs with less llvm-lines

dishmaker:
85 llvm-lines less

Before:
```
toml_edit$ cargo llvm-lines --release | head -n 5
  Lines                Copies              Function name
  -----                ------              -------------
  53438                1411                (TOTAL)
```

After:
```
toml_edit$ cargo llvm-lines --release | head -n 5
  Lines                Copies              Function name
  -----                ------              -------------
  53353                1408                (TOTAL)
```

Also, this removes `.collect()` in `table_enum.rs`.

### DIFF
--- a/crates/toml_edit/src/array.rs
+++ b/crates/toml_edit/src/array.rs
@@ -404,12 +404,10 @@ impl IntoIterator for Array {
     type IntoIter = ArrayIntoIter;
 
     fn into_iter(self) -> Self::IntoIter {
-        Box::new(
-            self.values
-                .into_iter()
-                .filter(|v| v.is_value())
-                .map(|v| v.into_value().unwrap()),
-        )
+        Box::new(self.values.into_iter().filter_map(|v| match v {
+            Item::Value(value) => Some(value),
+            _ => None,
+        }))
     }
 }
 
@@ -423,17 +421,16 @@ impl<'s> IntoIterator for &'s Array {
 }
 
 fn decorate_array(array: &mut Array) {
-    for (i, value) in array
-        .values
-        .iter_mut()
-        .filter_map(Item::as_value_mut)
-        .enumerate()
-    {
-        // [value1, value2, value3]
-        if i == 0 {
-            value.decorate(DEFAULT_LEADING_VALUE_DECOR.0, DEFAULT_LEADING_VALUE_DECOR.1);
-        } else {
-            value.decorate(DEFAULT_VALUE_DECOR.0, DEFAULT_VALUE_DECOR.1);
+    let mut first = true;
+    for value in array.values.iter_mut() {
+        if let Some(value) = Item::as_value_mut(value) {
+            // [value1, value2, value3]
+            if first {
+                value.decorate(DEFAULT_LEADING_VALUE_DECOR.0, DEFAULT_LEADING_VALUE_DECOR.1);
+            } else {
+                value.decorate(DEFAULT_VALUE_DECOR.0, DEFAULT_VALUE_DECOR.1);
+            }
+            first = false;
         }
     }
     // Since everything is now on the same line, remove trailing commas and whitespace.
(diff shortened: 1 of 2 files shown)


epage:
Could you help me understand what is motivating this?

Overall, these seem drops don't seem sufficient, aren't backed by real world numbers, and could easily be reverted back by any random change.

dishmaker:
Less llvm lines should compile faster.
`toml_edit` is used in the rust compiler, so this PR should speed up the development.

So far these PRs are additive, so small changes might add up :)

I will make some benchmarks in an hour.

dishmaker:
1.7% faster compile time when I merged all `toml_edit` llvm PRs.

  | after (all llvm PRs) | before (main)
-- | -- | --
average [s] | 2.856 | 2.905
stdev [s] | 0.011 | 0.012

<details>
  <summary>[Spoiler] Script compare.sh</summary>
  

https://github.com/dishmaker/toml/tree/dishmaker/all_llvms

```bash
echo > compare_mybranch.txt
echo > compare_master.txt

for run in {1..100}; do

git checkout dishmaker/all_llvms
cargo +nightly rustc --release --all-features -- -Z time-passes -C codegen-units=1 2>> compare_mybranch.txt

git checkout main
cargo +nightly rustc --release --all-features -- -Z time-passes -C codegen-units=1 2>> compare_master.txt

done

grep "total" compare_mybranch.txt ; echo ____master_main____ ; grep "total" compare_master.txt
```

</details>

epage:
In #1058 when I brought up the difference between debug and release, I had overlooked that you care about was compile times.  In that situation, debug builds are the more important one.  That said, I do hope  rust-lang/cargo#15931 will allow some optimizations in development builds.

It still feels hard justifying this.  These changes are making the code less idiomatic and could easily be reverted in a follow up, rather than focusing on improvements that are sustainable, like the `serde_core` change.  They are also small.  I know a lot of small improvements can make a difference but individually, they don't seem justified and its hard to weigh that against some theoretical perfect state, especially when we have to keep in mind that builds happen in parallel, reducing the benefits of small incremental gains.  If `toml` was a "shows up in every dep tree" dependency, like proc-macro deps, serde, or even further down the list like clap, then things change a bit.

crates/toml_edit/src/array.rs
```diff
@@ -404,12 +404,10 @@ impl IntoIterator for Array {
     type IntoIter = ArrayIntoIter;
 
     fn into_iter(self) -> Self::IntoIter {
-        Box::new(
-            self.values
-                .into_iter()
-                .filter(|v| v.is_value())
-                .map(|v| v.into_value().unwrap()),
```
dishmaker, line 411: So the removal of `.unwrap()` here isn't justified?

crates/toml_edit/src/de/table_enum.rs
```diff
@@ -65,7 +65,7 @@ impl<'de> serde_core::de::VariantAccess<'de> for TableEnumDeserializer {
         match self.value {
             crate::Item::ArrayOfTables(values) => {
                 let values_span = values.span();
-                let tuple_values = values.values.into_iter().collect::<Vec<_>>();
+                let tuple_values = values.values.clone();
```
dishmaker, line 68: Here I actually simplified the code.

dishmaker:
1.3% less llvm-lines for debug builds too.


```
toml_edit$ git checkout dishmaker/all_llvms
toml_edit$ cargo llvm-lines | head -n 3
  Lines                Copies              Function name
  -----                ------              -------------
  55865                2257                (TOTAL)
toml_edit$ git checkout main
toml_edit$ cargo llvm-lines | head -n 3
  Lines                Copies              Function name
  -----                ------              -------------
  56611                2296                (TOTAL)
```

dishmaker:
Entire example binary is also 0.3% smaller

```
toml_edit$ git checkout dishmaker/all_llvms
toml_edit$ cargo bloat --example visit | grep ".text section"

9.6% 100.0% 778.7KiB               .text section size, the file size is 7.9MiB


toml_edit$ git checkout main
toml_edit$ cargo bloat --example visit | grep ".text section"

9.6% 100.0% 781.2KiB               .text section size, the file size is 7.9MiB
```

PR #1069, opened 2025-11-19
Allow `InlineTable` insert API to accept `Into<Value>`

terror:
This diff updates `InlineTable::insert` to accept any `V: Into<Value>` just like `get_or_insert`, converting internally before storing. This lets callers pass primitive literals (&str, numbers, etc.) without manual Value construction while preserving existing behavior for Value. There are no behavior changes beyond the added flexibility.

### DIFF
--- a/crates/toml_edit/src/inline_table.rs
+++ b/crates/toml_edit/src/inline_table.rs
@@ -377,10 +377,10 @@ impl InlineTable {
     }
 
     /// Inserts a key-value pair into the map.
-    pub fn insert(&mut self, key: impl Into<String>, value: Value) -> Option<Value> {
+    pub fn insert<V: Into<Value>>(&mut self, key: impl Into<String>, value: V) -> Option<Value> {
         use indexmap::map::MutableEntryKey;
         let key = Key::new(key);
-        let value = Item::Value(value);
+        let value = Item::Value(value.into());
         match self.items.entry(key.clone()) {
             indexmap::map::Entry::Occupied(mut entry) => {
                 entry.key_mut().fmt();


crates/toml_edit/src/inline_table.rs
```diff
@@ -395,9 +395,9 @@ impl InlineTable {
     }
 
     /// Inserts a key-value pair into the map.
-    pub fn insert_formatted(&mut self, key: &Key, value: Value) -> Option<Value> {
+    pub fn insert_formatted<V: Into<Value>>(&mut self, key: &Key, value: V) -> Option<Value> {
```
epage: I believe the `_formatted` variants generally don't have conversions.  I don't remember how intentional the different choices are but I think it can make sense from the perspective of this call is about being explicit in formatting

terror: Yeah this makes sense, I didn't look too deeply into the intent behind the `_formatted` variants. I've removed this from the diff!

crates/toml_edit/src/inline_table.rs
epage, line 1: > [refactor: Allow InlineTable insert API to accept Into<Value>](https://github.com/toml-rs/toml/pull/1069/commits/86590e06f69be890fd8aa02f200cc7c4e9eb3f91)

For future reference, a `refactor` is technically a change that has no user-visible impact.  My running of `cargo release changes` would not highlight this commit for me to include in the changelog.  I know it is there and can watch for it so it isn't a blocker for this but wanted to give a heads up for anything in the future.

terror:
@epage Was there still a blocker for this? I'd be happy to rebase and amend the commit message if that's needed.

epage:
Every time I look at this, I go back and forth.  We don't really have a good principled reason for what we do.  I'll go ahead and merge and we can re-evaluate when we change up tables (https://github.com/toml-rs/toml/issues/762#issuecomment-2263660813).

epage:
The release was yanked and the change reverted because it broke type inference, see #1074

Issue #623, opened 2023-10-16
toml_edit feature request: Allow a table's `doc_position` to be discarded


TheSignPainter98:
I'm writing a formatter using `toml_edit`, and I'd like to sort tables at the top level, however those parsed from an existing TOML document have a non-`None` `doc_position` field, which seems to prevent `Document::sort_values`-like methods from repositioning them. Currently I'm using the following quite awkward workaround:

```rust
let mut elems = document.iter_mut().collect::<Vec<_>>();
elems.sort_by(|(k1, _), (k2, _)| k1.cmp(k2));
elems.iter_mut().enumerate().for_each(|(i, (k, v))| {
    if Some(table) = v.as_table_mut() {
        table.set_position(i);
    }
});
```

Is there currently a way to avoid this?

If not, would it be possible to allow `Document::sort_values` etc. to work by adding either:

- Some `Table::clear_position` method which sets a Table's doc_position to `None`, or
- New `sort_values`-like methods which don't take into account doc_position?

A disadvantage of allowing the position to be discarded would be that commented-out blocks of TOML could be re-ordered into unexpected places (as they are attached to some other item).

I'm happy to write a PR for this, but thought I'd post an issue first in case there's anything obvious I've missed

epage:
An alternative optin is for `Table::set_position` to accept `Option<usize>`.

It looks like we do stable sorts, so it should be reasonable to just do `table.set_position(0)` on top of the sorting you do.

TheSignPainter98:
I wasn't sure about the breaking change of moving `Table::set_position` to accept `Option<usize>` but all things being equal, this would be my preferred option. Should I open a PR for this?

Thanks for the suggestion, `table.set_position(0)` seems to work so I'll apply it to my stuff for now.

epage:
Unsure when/if we should next make a breaking change.  Especially as there is a viable workaround for you, I think I'd rather hold off.

TheSignPainter98:
Ah fair enough, thanks for the help!

nohajc:
Can we solve this with #1070?

I'm trying to merge two config files without any reordering. I simply want to preserve the order of insertion while reseting any position information from the source items before inserting them to the destination. Basically merging tables. The only alternative seems to be to recreate all items I need without the position property.

nohajc:
@epage Let me know if you need any more details. Thanks.

epage:
> I'm trying to merge two config files without any reordering. 

[...]

PR #1070, opened 2025-11-23
Table: add clear_position

nohajc:
This has already been discussed in #623. Unlike that case, I don't have any reasonable workaround.
My code combines two config files together, so I need a way to reset position before merging tables.

### DIFF
--- a/crates/toml_edit/src/table.rs
+++ b/crates/toml_edit/src/table.rs
@@ -241,6 +241,11 @@ impl Table {
         self.doc_position = Some(doc_position);
     }
 
+    /// Clears the position of the `Table` within the [`DocumentMut`][crate::DocumentMut].
+    pub fn clear_position(&mut self) {
+        self.doc_position = None;
+    }
+
     /// The position of the `Table` within the [`DocumentMut`][crate::DocumentMut].
     ///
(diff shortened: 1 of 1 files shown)


epage:
Our contrib guide asks for people to discuss things in issues before moving on to PRs.

Since I'm not seeing a reason for why your case is different than #623, I'm going to close this and suggest describing your use case in #623 for why we should re-evaluate opening the issue.

nohajc:
I'm sorry. I already mentioned my use case is merging configs and I don't want to do any sorting (which was the workaround in #623). If you need me to, I will copy-paste my comment to the issue.

nohajc:
I am also happy to provide examples if my use case is not clear.

But frankly, not having this API is kind of a glaring omission in an otherwise very useful library.

I understand breaking changes would be problematic, but this kind of approach seems perfectly fine to me.

epage:
As I said, this PR is not the place to discuss this.

nohajc:
Apologies. I just thought I'd save everyone's time. Haven't realized this is a strict policy.

epage:
It ends up bifurcating the conversation between the issue and the PR.  If we then don't move forward with this PR and someone else comes along, we then lose this conversations to help that conversation.

PR #1079, opened 2025-12-17
feat: Support TOML 1.1

### LINKED ISSUE #397: Support next TOML release

epage:
Fixes #397

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -980,7 +980,7 @@ dependencies = [
 
 [[package]]
 name = "toml_edit"
-version = "0.23.10+spec-1.0.0"
+version = "0.23.10+spec-1.1.0"
 dependencies = [
  "anstream",
  "anstyle",
--- a/crates/toml_edit/Cargo.toml
+++ b/crates/toml_edit/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "toml_edit"
-version = "0.23.10+spec-1.0.0"
+version = "0.23.10+spec-1.1.0"
 description = "Yet another format-preserving TOML parser."
 categories = ["encoding", "parser-implementations", "parsing", "config"]
 keywords = ["encoding", "toml"]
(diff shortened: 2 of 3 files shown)


crates/toml/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "toml"
-version = "0.9.9+spec-1.0.0"
+version = "0.9.9+spec-1.1.0"
```
epage, line 3: Blocked on the release of TOML 1.1: toml-lang/toml#928

epage: We should audit https://github.com/toml-rs/toml/issues?q=is%3Aissue%20state%3Aopen%20label%3AM-breaking-change before releasing

epage: #623 is the main candidate for `toml_edit` (the only package with a breaking change).  Others will require more investigation and more work to update to it.

Turbo87: just in case: those two versions numbers are equivalent and you won't be able to publish them aside each other

epage: Versions are bumped on release. I was just updating the advertised supported spec version.

Turbo87: alright, nevermind then :)

epage: To expand: Cargo patches require exact version matches, so if someone wants to take whats on github and try it out locally, they can't if I bump versions.  I might have broken that with adding metadata (haven't checked how strict cargo is) but that should rarely change.

PR #1094, opened 2026-02-03
feat: expose span() on TableLike

TyberiusPrime:
TableLike is missing the span() function that both Table and InlineTable have, 
this PR adds it.

I need to know this for my pretty error messages, and would like to be able to use the TableLike abstraction.

I notice that other functions like decor(), retain(), trailing(), trailing_comma() are also missing - so maybe this is by design? 
Or maybe the PR should be larger.

Thanks for all the hard work either way!

PS: Placing the couple of calls was easy. Finding where to add the right tests isn't :).

### DIFF
--- a/crates/toml_edit/src/inline_table.rs
+++ b/crates/toml_edit/src/inline_table.rs
@@ -609,6 +609,10 @@ impl TableLike for InlineTable {
     fn key_mut(&mut self, key: &str) -> Option<KeyMut<'_>> {
         self.key_mut(key)
     }
+
+    fn span(&self) -> Option<std::ops::Range<usize>> {
+        self.span()
+    }
 }
 
 // `{ key1 = value1, ... }`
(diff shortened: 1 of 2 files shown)


epage:
Note that we generally ask for Issues to discuss people's problem and agree to a solution before moving on to PRs ([contrib](https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#pull-requests)).  One reason is that it keeps those conversations in one place for easier reference in the future.

> TableLike is missing the span() function that both Table and InlineTable have,
this PR adds it.
>
> I notice that other functions like decor(), retain(), trailing(), trailing_comma() are also missing - so maybe this is by design?
Or maybe the PR should be larger.

These all have different meanings between a `Table` and an `InlineTable` and so don't make sense to expose on `TableLike`.

TyberiusPrime:
> Note that we generally ask for Issues to discuss people's problem and agree to a solution before moving on to PRs ([contrib](https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#pull-requests)). One reason is that it keeps those conversations in one place for easier reference in the future.

My apologies, the contrib page suggested that this preference was to avoid rework on PRs, and well, it's a 5 very trivial lines, and I'll gladly take the risk of having to rework them. Keeping conversations in one place is not mentioned in the contrib page, perhaps consider adding it.

I'll open an issue for the 'missing' non-span methods.

PR #1113, opened 2026-03-02
Add support for big integers

threema-mike:
Hello toml_edit team,
For context, I am using toml_edit to write out TOML files with comments, generated from annotated Rust structs.
Since we want to be able to also use big integers for certain config options, and the `toml` crate supports big ints, I figured I'll try to get it working for `toml_edit` too.
The initial PR is a discussion basis, I assume that you might want to feature-gate this or follow a different approach. I am open to discuss and adapt to the needs.

For my use-case those changes make it work for serialising and deserialising, this is therefore a working proof of concept.

### DIFF
--- a/crates/toml_edit/src/de/value.rs
+++ b/crates/toml_edit/src/de/value.rs
@@ -4,6 +4,7 @@ use crate::de::ArrayDeserializer;
 use crate::de::DatetimeDeserializer;
 use crate::de::Error;
 use crate::de::TableDeserializer;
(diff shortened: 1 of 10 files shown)


epage:
> The initial PR is a discussion basis, I assume that you might want to feature-gate this or follow a different approach. I am open to discuss and adapt to the needs.

We ask that design discussions happen in Issues, not PRs, see https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#pull-requests

PR #1122, opened 2026-03-21
feat(edit): Support insert() on ArrayOfTables

dimbleby:
Looks like a gap on the API?


epage:
For reference, compared to `Array`, we are missing
- `replace`
- `sort_by`
- `sort_by_key`

dimbleby:
> For reference, compared to Array, we are missing ...

I find that I actually have a use for `replace()`, may follow up with that one

PR #1123, opened 2026-03-21
fix(edit): extract key comment before table header bracket

### LINKED ISSUE #691: Replacing an `InlineTable` with a comment with a `Table` results in invalid syntax
```rust
use std::str::FromStr;

use toml_edit::{Document, Item, Table};

fn main() {
    let mut val = Document::from_str(
        r#"
    # hello i'm a comment
    foo = { bar = 1 }
    "#,
    )
    .unwrap();
    println!("before: {}", val.to_string());
    let mut new_foo = Table::new();
    new_foo.insert("bar", Item::Value(1.into()));
    new_foo.insert("baz", Item::Value(2.into()));
    *val.get_mut("foo").unwrap() = Item::Table(new_foo);
    println!("after: {}", val.to_string());
}
```
Expected output:
```
before:
    # hello i'm a comment
    foo = { bar = 1 }

after:
    # hello i'm a comment
    [foo]
    bar = 1
    baz = 2
```
Actual output:
```
before:
    # hello i'm a comment
    foo = { bar = 1 }

after: [
    # hello i'm a comment
    foo ]
bar = 1
baz = 2
```

dimbleby:
A try at fixing #691.  I am in two minds about whether this is more of a sticking plaster than a real "fix".  But it is relatively tightly scoped - and seems to work. 

Fixes #691

### DIFF
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -8,7 +8,7 @@ use toml_writer::TomlWrite as _;
 use crate::DocumentMut;
 use crate::inline_table::DEFAULT_INLINE_KEY_DECOR;
 use crate::key::Key;
-use crate::repr::{Formatted, Repr, ValueRepr};
+use crate::repr::{Decor, Formatted, Repr, ValueRepr};
 use crate::table::{
     DEFAULT_KEY_DECOR, DEFAULT_KEY_PATH_DECOR, DEFAULT_ROOT_DECOR, DEFAULT_TABLE_DECOR,
 };
@@ -37,8 +37,8 @@ fn encode_key_path(
     mut buf: &mut dyn Write,
     input: Option<&str>,
     default_decor: (&str, &str),
+    leaf_decor: &Decor,
 ) -> Result {
-    let leaf_decor = this.last().expect("always at least one key").leaf_decor();
     for (i, key) in this.iter().enumerate() {
         let dotted_decor = key.dotted_decor();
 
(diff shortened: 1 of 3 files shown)


crates/toml_edit/src/encode.rs
```diff
@@ -37,16 +37,21 @@ fn encode_key_path(
     mut buf: &mut dyn Write,
     input: Option<&str>,
     default_decor: (&str, &str),
+    // Leaf decor to encode around the key path. When the leaf decor came from a
+    // key-value pair context (prefix already written before the opening bracket,
+    // suffix not meaningful inside `[...]`), pass `None` to skip it.
+    leaf_decor: Option<&Decor>,
```
epage: If we move forward with this, I think passing in a `&Decor::default()` would be more correct than `None` as it would cause the default decor to be respected.

crates/toml_edit/tests/testsuite/edit.rs
```diff
@@ -1681,3 +1681,53 @@ tool = { typst-test.tests = "tests" }
 "#]]
     );
 }
+
+#[test]
+fn inline_table_to_table_with_comment() {
```
epage, line 1788: We generally ask that tests be added in a commit before a fix with them passing, reproducing the undesired behavior.  This helps illustrate the problem for anyone looking at this PR.

See also https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#process

crates/toml_edit/src/encode.rs
```diff
@@ -296,9 +315,17 @@ fn visit_table(
...
         };
+        let extracted = leaf_decor_before_bracket(path, input);
+        if let Some(leaf_decor) = extracted {
+            leaf_decor.prefix_encode(buf, input, DEFAULT_KEY_PATH_DECOR.0)?;
+        }
         table.decor.prefix_encode(buf, input, default_decor.0)?;
+        let leaf_decor = match extracted {
+            Some(_) => None,
+            None => path.last().map(|key| key.leaf_decor()),
+        };
```
epage: I think I'd change this to where a function like `leaf_decor_before_bracket` returns
- `Option<Decor>` for before the table header
- `Decor` for the key in the table header

crates/toml_edit/tests/testsuite/edit.rs
dimbleby: gonna need to resolve conflicts with my other PR anyway - both have tests at the end of the same file - so can do that.

dimbleby:
rearranged, hopefully along the expected lines.  I introduced `Decor::EMPTY` so as to avoid the need to clone.

PR #1127, opened 2026-03-23
feat(edit): Support replace() on ArrayOfTables

dimbleby:
As noted in #1122 - ArrayOfTables is missing `replace()` and `sort_by()` and `sort_by_key()` compared to the Array.

I also wrote the naive implementations for the sorts.  But even when the internal `Vec` is successfully sorted, the encode sorts by `position()` regardless - so extra effort is needed to make that do something sensible.

I am sure it cannot be very hard to make this work - but I have a use for `replace` and not for `sort`, so I prefer to leave that for someone who is more motivated.

### DIFF
--- a/crates/toml_edit/src/array_of_tables.rs
+++ b/crates/toml_edit/src/array_of_tables.rs
@@ -99,6 +99,29 @@ impl ArrayOfTables {
         self.values.insert(index, Item::Table(table));
     }
 
+    /// Replaces a table at the given index within the array, returning the old table.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index >= len`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use toml_edit::{ArrayOfTables, Table, value};
+    ///
+    /// let mut arr = ArrayOfTables::new();
+    /// arr.push(Table::from_iter([("name", value("apple"))]));
+    ///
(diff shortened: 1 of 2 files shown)


crates/toml_edit/src/array_of_tables.rs
```diff
@@ -99,6 +99,57 @@ impl ArrayOfTables {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// let mut arr = toml_edit::ArrayOfTables::new();
+    /// arr.push(toml_edit::Table::new());
+    /// arr.push(toml_edit::Table::new());
+    ///
+    /// arr.replace_formatted(0, toml_edit::Table::new());
+    /// ```
+    pub fn replace_formatted(&mut self, index: usize, table: Table) -> Table {
```
epage: So far, `ArrayOfTables` hasn't had a split plain/formatted but every method is formatted.

Is there a reason to break from that here?

dimbleby: I was just copying what was on the `Array`, happy not to do so if that was the wrong call

dimbleby: yes, I see - the `Array` has two versions of everything, but the `ArrayOfTables` does not.  I will change it

epage:
Sorry, it looks like I forgot to release this before

PR #1130, opened 2026-03-29
fix(toml): convert recursion to iteration

dimbleby:
Avoiding a stack overflow on a long run of redundant signs.

Fixes #1129

Issue #1129, opened 2026-03-29
stack-overflow in decode_sign_prefix


van-ema:
Hello.

During fuzzing campaign I found that `toml_parser` recursively strips redundant leading numeric signs in `decode_sign_prefix` around line 170. A sufficiently long run of `+` or `-` characters causes unbounded recursion and ends in stack overflow.

## Reproduce

Create input file with `8100` `-` characters after `a=`.

```
python3 - <<'PY' >/tmp/poc_8100_signs.toml
  print("a=" + "-" * 8100, end="")
  PY
```

POC:
```
  mkdir -p /tmp/toml-stack-repro/src

  cat >/tmp/toml-stack-repro/Cargo.toml <<EOF
  [package]
  name = "toml-stack-repro"
  version = "0.1.0"
  edition = "2021"

  [dependencies]
  toml = { path = "$PWD/crates/toml" }
  EOF
```

```
  cat >/tmp/toml-stack-repro/src/main.rs <<'EOF'
  fn main() {
      let path = std::env::args().nth(1).expect("need input path");
      let input = std::fs::read_to_string(path).unwrap();
      let _ = toml::from_str::<toml::Table>(&input);
  }
  EOF
```

Run with asan:

```
  ASAN_OPTIONS=detect_leaks=0:abort_on_error=1:symbolize=0 \
  RUSTFLAGS="-Zsanitizer=address" \
  cargo +nightly run --manifest-path /tmp/toml-stack-repro/Cargo.toml -- /tmp/poc_8100_signs.toml
```

PR #1131, opened 2026-03-29
fix(parser): Fix stack overflow in value() on repeated '='

dimbleby:
Similar to #1129 and #1130

Rewrite the extra '=' handling in value() to iteratively skip redundant '=' tokens in a loop instead of recursing, preventing stack overflow on long runs of '=' characters. Add regression test.

### DIFF
--- a/crates/toml_parser/src/parser/document.rs
+++ b/crates/toml_parser/src/parser/document.rs
@@ -626,27 +626,40 @@ fn opt_dot_keys(
 /// val = string / boolean / array / inline-table / date-time / float / integer
 /// ```
 fn value(tokens: &mut Stream<'_>, receiver: &mut dyn EventReceiver, error: &mut dyn ErrorSink) {
-    let Some(current_token) = tokens.next_token() else {
-        let previous_span = tokens
-            .previous_tokens()
-            .find(|t| {
-                !matches!(
-                    t.kind(),
-                    TokenKind::Whitespace
-                        | TokenKind::Comment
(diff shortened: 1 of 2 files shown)


crates/toml_parser/src/parser/document.rs
```diff
@@ -626,27 +626,41 @@ fn opt_dot_keys(
...
-                    TokenKind::Whitespace
-                        | TokenKind::Comment
-                        | TokenKind::Newline
-                        | TokenKind::Eof
-                )
-            })
-            .map(|t| t.span())
-            .unwrap_or_default();
+    // Loop to iteratively skip extra `=` tokens, avoiding stack overflow on
+    // long runs.
```
epage: imo this comment is what motivates the change and not really needed to live in the code long term.  The logic covers that this needs to iterate somehow and the test covers one requirement on how that is done.

crates/toml_parser/src/parser/document.rs
```diff
@@ -706,6 +710,8 @@ fn value(tokens: &mut Stream<'_>, receiver: &mut dyn EventReceiver, error: &mut
         | TokenKind::Atom => {
             on_scalar(tokens, current_token, receiver, error);
         }
+        // Handled by the while loop above
+        TokenKind::Equals => unreachable!(),
```
epage: We generally keep the variants in order, so please move this back to the original location

epage:
Thanks!

Issue #1120, opened 2026-03-09
Error highlight doesn't handle multi-byte characters correctly


MKindberg:
If an error occurs when parsing a string containing utf-8 characters made up of multiple bytes like 'ö', whitespace after the string will be highlighted in addition to the string, for example

```Rust
use serde::Deserialize;

#[derive(Deserialize, Debug)]
struct Config {
    num: u32,
}

fn main() {
    if let Err(e) = toml::from_str::<Config>("num = 'åäö'") {
        println!("{}", e);
    }
}
```
Will give the following output
```text
TOML parse error at line 1, column 7
  |
1 | num = 'åäö'
  |       ^^^^^^^^
invalid type: string "åäö", expected u32
```

epage:
This is likely something we won't fix.

We need to balance the quality of our error messages with build times and user complexity because many users would likely want to use an alternative implementation, like https://crates.io/crates/annotate-snippets.  Correctly knowing the display width of characters on a terminal is tricky and would likely require pulling in other dependencies.

PR #1132, opened 2026-03-29
fix(parser): Fix non-ASCII span in decode_unquoted_key

dimbleby:
Iterate by char (char_indices) instead of by byte, so multi-byte UTF-8 characters produce one error with a correct span rather than one error per byte at potentially mid-codepoint offsets. Also fixes empty spans (i..i) to properly cover the character (i..i+len). Add regression test.

### DIFF
--- a/crates/toml_parser/src/decoder/string.rs
+++ b/crates/toml_parser/src/decoder/string.rs
@@ -702,8 +702,8 @@ pub(crate) fn decode_unquoted_key<'i>(
         );
     }
 
-    for (i, b) in s.as_bytes().iter().enumerate() {
-        if !UNQUOTED_CHAR.contains_token(b) {
+    for (i, ch) in s.char_indices() {
+        if !ch.is_ascii() || !UNQUOTED_CHAR.contains_token(&(ch as u8)) {
             error.report_error(
                 ParseError::new("invalid unquoted key")
                     .with_context(Span::new_unchecked(0, s.len()))
@@ -713,7 +713,7 @@ pub(crate) fn decode_unquoted_key<'i>(
                         Expected::Literal("-"),
                         Expected::Literal("_"),
                     ])
-                    .with_unexpected(Span::new_unchecked(i, i)),
+                    .with_unexpected(Span::new_unchecked(i, i + ch.len_utf8())),
             );
         }
     }
--- a/crates/toml/src/de/error.rs
+++ b/crates/toml/src/de/error.rs
@@ -127,6 +127,11 @@ impl core::fmt::Display for Error {
             let highlight_len = span.end - span.start;
             // Allow highlight to go one past the line
             let highlight_len = highlight_len.min(content.len().saturating_sub(column));
+            // Convert byte length to character count to match the char-based column
+            let highlight_len = input
+                .get(span.start..span.start + highlight_len)
+                .map(|s| s.chars().count())
+                .unwrap_or(highlight_len);
 
             writeln!(f, "TOML parse error at line {line_num}, column {col_num}")?;
             //   |
--- a/crates/toml/tests/snapshots/invalid/encoding/ideographic-space.stderr
+++ b/crates/toml/tests/snapshots/invalid/encoding/ideographic-space.stderr
@@ -3,17 +3,3 @@ TOML parse error at line 2, column 1
 2 | 　foo = "bar"
   | ^
 invalid unquoted key, expected letters, numbers, `-`, `_`
-
----
-TOML parse error at line 2, column 2
-  |
-2 | 　foo = "bar"
-  |  ^
-invalid unquoted key, expected letters, numbers, `-`, `_`
-
----
-TOML parse error at line 2, column 1
-  |
-2 | 　foo = "bar"
-  | ^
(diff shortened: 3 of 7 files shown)


dimbleby:
NB a proper fix to output the correct number of "^^" in the errors - see in particular the ideographic-space case - requires calculating display width.  Probably by taking a dependency on [unicode-width](https://crates.io/crates/unicode-width).  

Seems like the right action - but still not something I would do without encouragement.

dimbleby:
I have just realised that this is related to #1120, in particular the suggestion to bring in unicode-width would fix it - but is sort-of discouraged at [this comment](https://github.com/toml-rs/toml/issues/1120#issuecomment-4023554931)

`.chars().count()` is a step in the right direction, `.width()` would be better.

epage:
> Iterate by char (char_indices) instead of by byte, so multi-byte UTF-8 characters produce one error with a correct span rather than one error per byte at potentially mid-codepoint offsets. **Also** fixes empty spans (i..i) to properly cover the character (i..i+len). Add regression test.

Note my emphasis.  Our [contrib doc](https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#process) asks for commits to be atomic.

crates/toml_parser/src/decoder/string.rs
```diff
@@ -702,8 +702,8 @@ pub(crate) fn decode_unquoted_key<'i>(
         );
     }
 
-    for (i, b) in s.as_bytes().iter().enumerate() {
-        if !UNQUOTED_CHAR.contains_token(b) {
+    for (i, ch) in s.char_indices() {
+        if !ch.is_ascii() || !UNQUOTED_CHAR.contains_token(&(ch as u8)) {
```
epage, line 706: Iterating by `char` is slower because it needs to check the character boundaries and decode the content to a `u32`.

One option might be to batch invalid characters into the errors:
```rust
let mut span = None;
for i in s.as_bytes().iter().enumerate().filter(|(_, b)| !UNQUOTED_CHAR.contains_token(b)) {
    if let Some((start, end)) = span {
        if i == end {        
            span = Some((start, i + 1));
        } else {
            // .. error
            span = Some((i, i + 1));
        }
    } else {
        span = Some((i, i + 1));
    }
}
if let Some(span) = span {
    // .. error
}
```

crates/toml_parser/tests/testsuite/parse_simple_key.rs
```diff
@@ -34,3 +34,8 @@ fn simple_key_string() {
 fn simple_key_string_literal() {
     t(r"'hello\n '", file![_].raw());
 }
+
+#[test]
+fn simple_key_unquoted_non_ascii() {
+    t("é", file![_].raw());
+}
```
epage: We ask that tests get added in a commit before fixes, with them passing, showing the current behavior.

crates/toml_parser/src/decoder/string.rs
dimbleby: I dont follow the suggestion.  

I do not think that any approach that insists on spans being one byte wide can be correct.  This will later create one-byte slices into utf8-strings in a way that is certainly undefined behaviour.  

The current code creates zero length slices in the middle of utf8 characters.  It is unclear to me whether this already is undefined behaviour, or only skirting very very close to it.

We have to be at least somewhat utf8-character-aware, so that we can create a legal slice of length `ch.len_utf8()`

crates/toml_parser/tests/testsuite/parse_simple_key.rs
dimbleby: I removed the new testcase, seems like there are already enough cases including non-ascii characters, and which show the changed behaviour

dimbleby:
> > Iterate by char (char_indices) instead of by byte, so multi-byte UTF-8 characters produce one error with a correct span rather than one error per byte at potentially mid-codepoint offsets. **Also** fixes empty spans (i..i) to properly cover the character (i..i+len). Add regression test.
> 
> Note my emphasis. Our [contrib doc](https://github.com/toml-rs/toml/blob/main/CONTRIBUTING.md#process) asks for commits to be atomic.

Perhaps the bot has chosen the wrong word in the commit message here.  It all really is the same fix: the code around spans mishandles utf8 characters.  I do not think it makes sense to take the changes in any one place without also taking the others.

crates/toml_parser/src/decoder/string.rs
dimbleby: my benchmarking suggests that an up-front `s.is_ascii()` check is practically free - actually somewhat _better_ than free for strings up to 64K.  (Presumably something helpful is happening with L1 cache? not sure).

So I suppose we could go:
```
if ascii
  the loop we currently have
else
  the utf8-careful loop
```
?

epage: > I dont follow the suggestion.

I've merged it as #1138.  Feel free to play with it and give me your thoughts.

epage:
> Perhaps the bot has chosen the wrong word in the commit message here.

If you mean that any or all of this was done by an LLM, keep in mind that you are responsible for being able to understand and justify your PRs.

PR #1133, opened 2026-03-30
Optimisations

dimbleby:
A couple of small optimisations:
- avoid unnecessary path cloning in `get_values()`
- avoid needlessly collecting one already-owned `Vec` into another 

The second seems easy, the first is maybe a little bit on the clever side for modest gain.  

If you want both, or neither, thats easy: merge or reject.  If you want one but not the other let me know.

### DIFF
--- a/crates/toml_edit/src/de/table_enum.rs
+++ b/crates/toml_edit/src/de/table_enum.rs
@@ -65,7 +65,7 @@ impl<'de> serde_core::de::VariantAccess<'de> for TableEnumDeserializer {
         match self.value {
             crate::Item::ArrayOfTables(values) => {
                 let values_span = values.span();
-                let tuple_values = values.values.into_iter().collect::<Vec<_>>();
+                let tuple_values = values.values;
 
                 if tuple_values.len() == len {
                     serde_core::de::Deserializer::deserialize_seq(
@@ -81,7 +81,7 @@ impl<'de> serde_core::de::VariantAccess<'de> for TableEnumDeserializer {
             }
             crate::Item::Value(crate::Value::Array(values)) => {
                 let values_span = values.span();
-                let tuple_values = values.values.into_iter().collect::<Vec<_>>();
+                let tuple_values = values.values;
 
                 if tuple_values.len() == len {
                     serde_core::de::Deserializer::deserialize_seq(


crates/toml_edit/src/inline_table.rs
```diff
@@ -53,31 +53,31 @@ impl InlineTable {
...
         values: &mut Vec<(Vec<&'s Key>, &'s Value)>,
     ) {
         for (key, value) in self.items.iter() {
-            let mut path = parent.to_vec();
-            path.push(key);
+            parent.push(key);
             match value {
                 Item::Value(Value::InlineTable(table)) if table.is_dotted() => {
-                    table.append_values(&path, values);
+                    table.append_values(parent, values);
```
epage: This is no longer a parent at this point without convoluting the meaning

crates/toml_edit/src/inline_table.rs
```diff
@@ -53,31 +53,31 @@ impl InlineTable {
...
-        self.append_values(&root, &mut values);
+        let mut root = Vec::new();
+        self.append_values(&mut root, &mut values);
         values
     }
 
     pub(crate) fn append_values<'s>(
         &'s self,
-        parent: &[&'s Key],
+        parent: &mut Vec<&'s Key>,
```
epage: If we do this, it is unusual enough that this will need documentation

crates/toml_edit/src/table.rs
```diff
@@ -75,7 +75,7 @@ impl Table {
     ///
     /// `path` is the path so far in the walk through the tree. This function may add to it as it
     /// recurses through the tree, but on return it must be the same as it was on entry.
-    fn append_values<'s>(
+    pub(crate) fn append_values<'s>(
```
epage: These two functions don't look equivalent, see the `Item::Table` variant.  The fact that no tests failed though is interesting.

crates/toml_edit/src/inline_table.rs
```diff
@@ -53,31 +53,35 @@ impl InlineTable {
...
-        self.append_values(&root, &mut values);
+        let mut root = Vec::new();
+        self.append_values(&mut root, &mut values);
         values
     }
 
+    /// Helper for `get_values()`.
+    ///
+    /// `path` is the path so far in the walk through the tree. This function may add to it as it
+    /// recurses through the tree, but on return it must be the same as it was on entry.
```
epage: The current wording focuses on the implementation, making it harder for callers to identify what they need to know.

What about:

> `path` is the parent for this table.  `path` is mutable to reuse allocations but no mutations should be observable.

crates/toml_edit/src/table.rs
dimbleby: oh you mean I did not spot the `if table.is_dotted()`?

Indeed that was an oversight - encouraged by no tests failing.  I'll reinstate the not-quite-duplicate, probably got carried away.

crates/toml_edit/src/inline_table.rs
dimbleby: if you're happy, I'm happy

epage:
Thanks!

PR #1139, opened 2026-04-01
fix(parser): use unicode width in error reporting

### LINKED ISSUE #1120: Error highlight doesn't handle multi-byte characters correctly

dimbleby:
Recognising that you may not want the dependency - I thought I would at least give you a pull request you can turn down.

Using `unicode-width`, we can highlight the right number of columns, instead of just counting the bytes.

Fixes #1120

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -967,6 +967,7 @@ dependencies = [
  "toml_datetime",
  "toml_parser",
  "toml_writer",
+ "unicode-width",
  "walkdir",
  "winnow",
 ]
@@ -1048,6 +1049,7 @@ dependencies = [
  "toml_datetime",
  "toml_parser",
  "toml_writer",
+ "unicode-width",
  "walkdir",
  "winnow",
 ]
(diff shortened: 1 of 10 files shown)


dimbleby:
spell check failures in CI are in main, introduced at cc37615fc8d9213594250043e0e3b6cf73ed1e69

epage:
A decision is needed in #1120 before we are ready to move onto a PR.

dimbleby:
Sure, please discuss!

PR #1171, opened 2026-06-08
Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, these clones can be safely removed.

CC @nunoplopes


epage:
Note that removing these clones will make no difference in performance.

PR #1192, opened 2026-07-25
docs(toml_edit): document trailing newline limitation

### LINKED ISSUE #1158: Preserve absence of final newline when writing? Or document not doing so?
Hi!

In brief: Currently, `toml_edit` adds a `\n` after the text if there is none. If this is deemed to fall outside the "preserving […] spaces" description in the docs, perhaps it would be a good idea to either.

- Document the decision/behavior; or
- Preserve the absence of such a newline.

The second is of course a larger (mildly breaking) change. Also, the spec does say that there must be a newline after each key–value pair (though this is perhaps a bit imprecise, since comments are clearly permitted on the same line, anyway). However, since the parser accepts such a file as input, I guess the validity argument isn't that strong.

From what I can see, the reason this happens is that the parsing splits into lines, removing the line endings, and then there's an unconditional `writeln!` at the output end. I suppose to change this, one would need (e.g.) a boolean field representing whether there should be a newline at the end?

And I'm not saying this would necessarily be worthwhile. The reason I came across it was just that I had tests involving one-liner TOML documents which then hadd a `\n` appended, which surprised me a bit.

Now, there would presumably still be *some* newline normalization. That is, preserving CRLF is a much larger step to take. So … suppose the most straightforward thing would be to add a mention in the docs about newlines being normalized, including ensuring that there is one at the end?

yashkuceriya:
Closes #1158.

## Summary

Document that `toml_edit` normalizes output to end with a trailing newline and does not preserve whether the input omitted it.

Add the note to both the crate README and the crate-level docs alongside the existing limitations section.

## Repro

```rust
let input = "answer = 42";
let doc = input.parse::<toml_edit::DocumentMut>()?;
assert_eq!(doc.to_string(), "answer = 42\n");
```

## Validation

- `cargo fmt --all --check`
- `cargo test -p toml_edit`
- `cargo clippy -p toml_edit --lib -- -D warnings`
- `/private/tmp/prek-venv/bin/prek run --all-files`

### DIFF
--- a/crates/toml_edit/README.md
+++ b/crates/toml_edit/README.md
@@ -41,6 +41,7 @@ c = { d = "hello" }
 Things it does not preserve:
 
 * Order of dotted keys, see [issue](https://github.com/toml-rs/toml/issues/163).
+* Lack of trailing newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
 
 ## License
 
--- a/crates/toml_edit/src/lib.rs
+++ b/crates/toml_edit/src/lib.rs
@@ -66,6 +66,7 @@
 //! Things it does not preserve:
 //!
 //! * Order of dotted keys, see [issue](https://github.com/toml-rs/toml/issues/163).
+//! * Lack of trailing newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
 //!
 //! [`toml`]: https://docs.rs/toml/latest/toml/
 


crates/toml_edit/src/lib.rs
```diff
@@ -66,6 +66,7 @@
 //! Things it does not preserve:
 //!
 //! * Order of dotted keys, see [issue](https://github.com/toml-rs/toml/issues/163).
+//! * Whether the original document omitted a trailing newline. Output is normalized to end with a newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
```
epage:
```suggestion
//! * Lack of trailing newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
```

crates/toml_edit/README.md
```diff
@@ -41,6 +41,7 @@ c = { d = "hello" }
 Things it does not preserve:
 
 * Order of dotted keys, see [issue](https://github.com/toml-rs/toml/issues/163).
+* Whether the original document omitted a trailing newline. Output is normalized to end with a newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
```
epage:
```suggestion
* Lack of trailing newline, see [issue](https://github.com/toml-rs/toml/issues/1158).
```

epage:
Note that our contrib guide asks for atomic commits that reflect how a change should be reviewed and merged and not how it was developed.

Please squash your commits.

epage:
Thanks!

PR #1194, opened 2026-07-28
fix(toml): preserve datetimes when deserializing Value

### LINKED ISSUE #440: Explicit Deserialization of TOML Datetimes between structs and `Value` fail
**I tried this code**: [Rust Playground Link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aef0d73224738c3ce177295b69db860f)

**I expected this to happen**: both the first "full" deserialization and the second "partial" deserialization behave the same.

**Instead this happened**: error `invalid type: string \"[...]", expected a TOML datetime`

Possibly related to #333 very intdirectly? At least it seems related to the weird way that Datetime objects in TOML are handled.

ahmojo:
Closes #440.

## Summary

Fix explicit deserialization of TOML datetimes from `toml::Value` without
changing generic cross-format conversions.

`toml::value::Datetime` asks Serde for TOML's private datetime struct. The
legacy `toml::Value` deserializer forwarded `deserialize_struct` to
`deserialize_any`, where `Value::Datetime` is represented as a string. Typed
conversions therefore failed with:

```text
invalid type: string "1979-05-27T07:32:00Z", expected a TOML datetime
```

## What changed

- Handle TOML's private datetime struct in `Value::deserialize_struct`.
- Forward `Table::deserialize_struct` to the `Value` implementation.
- Keep `deserialize_any` unchanged so generic JSON conversion still produces
  a string.
- Add synchronized regression coverage to the `toml` and `toml_edit` Serde
  test suites.
- No new dependencies.

## Behavioral impact

`Value::try_into::<toml::value::Datetime>()` now succeeds for
`Value::Datetime`, while converting the same value into `serde_json::Value`
continues to produce a JSON string.

Other `toml::Value` variants and ordinary struct deserialization are
unchanged.

## Scope

The separate `Value::try_from` serialization behavior mentioned later in
#440 uses a different serializer path and is intentionally unchanged.

## Validation

- [x] `cargo +1.97.0 test -p toml --test serde general::deserialize_datetime_from_value_issue_440 -- --exact --nocapture`
- [x] `cargo +1.97.0 test -p toml_edit --features serde --test serde general::deserialize_datetime_from_value_issue_440 -- --exact --nocapture`
- [x] `cargo +1.97.0 test -p toml --test serde` (173 passed)
- [x] `cargo +1.97.0 test -p toml_edit --features serde --test serde` (163 passed)
- [x] `cargo +1.97.0 check -p toml --no-default-features --features serde --locked`
- [x] `cargo +1.97.0 clippy -p toml -p toml_edit --all-features --all-targets -- -D warnings`
- [x] `cargo +1.97.0 fmt --all --check`

### DIFF
--- a/crates/toml/src/value.rs
+++ b/crates/toml/src/value.rs
@@ -608,9 +608,26 @@ impl<'de> de::Deserializer<'de> for Value {
         visitor.visit_newtype_struct(self)
     }
 
+    fn deserialize_struct<V>(
+        self,
+        name: &'static str,
+        _fields: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, crate::de::Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        match (toml_datetime::de::is_datetime(name), self) {
+            (true, Self::Datetime(v)) => {
+                visitor.visit_map(toml_datetime::de::DatetimeDeserializer::new(v))
+            }
+            (_, value) => value.deserialize_any(visitor),
+        }
+    }
+
     serde_core::forward_to_deserialize_any! {
         bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string unit seq
-        bytes byte_buf map unit_struct tuple_struct struct
+        bytes byte_buf map unit_struct tuple_struct
         tuple ignored_any identifier
     }
 }
--- a/crates/toml/tests/serde/general.rs
+++ b/crates/toml/tests/serde/general.rs
@@ -1492,6 +1492,25 @@ fn table_type_enum_regression_issue_388() {
 }
 
 #[test]
+fn deserialize_datetime_from_value_issue_440() {
+    let input = "1979-05-27T07:32:00Z";
+    let value = crate::value_from_str::<crate::SerdeValue>(input).unwrap();
+
+    let json = value.clone().try_into::<serde_json::Value>().unwrap();
+    assert_eq!(json, serde_json::Value::String(input.to_owned()));
+
+    let error = value.try_into::<crate::Datetime>().unwrap_err();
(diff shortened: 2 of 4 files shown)


crates/toml/tests/serde/general.rs
```diff
@@ -1491,6 +1491,18 @@ fn table_type_enum_regression_issue_388() {
     assert!(crate::from_str::<DataFile>(inline_table).is_ok());
 }
 
+#[test]
+fn deserialize_datetime_from_value_issue_440() {
```
epage, line 1495: Our contrib guide asks for tests to be added before the change, reproducing the bad behavior.

See also https://epage.github.io/dev/pr-style/#c-split

epage, line 1495: All `serde` tests need to synchronized with `toml_edit`s `serde` tests

crates/toml/src/value.rs
```diff
@@ -527,7 +527,7 @@ impl<'de> de::Deserializer<'de> for Value {
             Self::Integer(n) => visitor.visit_i64(n),
             Self::Float(n) => visitor.visit_f64(n),
             Self::String(v) => visitor.visit_string(v),
-            Self::Datetime(v) => visitor.visit_string(v.to_string()),
+            Self::Datetime(v) => visitor.visit_map(toml_datetime::de::DatetimeDeserializer::new(v)),
```
epage: iirc this can be used to convert between formats. How does this interact with deserializing into json?

ahmojo: Thanks for pointing that out. With the revised fix, `toml::Value::Datetime` stays a JSON string; only TOML datetime targets receive the private map.

PR #1195, opened 2026-07-29
fix(edit): preserve tables when displaying an owned Document

### LINKED ISSUE #1008: `Display` implementation for `toml_edit::Document` is broken
The code below prints an empty string.
```rust
fn main() {
    let toml = r##"
[project]
x = "y"
    "##;

    let doc = toml.parse::<toml_edit::Document<String>>().unwrap();
    println!("{}", doc.to_string());
}
```

If I remove the `project` header it seems to print the key and value pair.

hdimer:
## Summary

`toml_edit::Document<String>::to_string()` returns an empty string when the document contains a `[table]` header (#1008):

```rust
let doc = "\n[project]\nx = \"y\"\n".parse::<toml_edit::Document<String>>().unwrap();
assert_eq!(doc.to_string(), ""); // before this PR
```

`DocumentMut` round-trips the same input correctly.

## Root cause

`Document<S>` has no `Display` impl of its own; it only `Deref`s to `Table`, so `to_string()` resolves to `<Table as Display>`, which renders the root table's *direct* items but not the `[table]` headers of nested tables. The document-level logic that emits those headers (`visit_nested_tables` + sorting by position) lives solely in `impl Display for DocumentMut`. So a document whose content lives under a header renders as nothing.

## Fix

Extract the document-encoding body into a shared `encode_document(buf, input, decor, root, trailing)` helper and give `Document<S>` its own `Display`. Because `Document<S>` keeps the parsed tree spanned against its retained source, it passes `Some(self.raw())` so the decor/reprs resolve against that source; `DocumentMut` is despanned and keeps passing `None`. `DocumentMut`'s output is byte-for-byte unchanged.

`Document<String>::to_string()` now losslessly round-trips its input and matches `DocumentMut`.

## Tests

Added a regression test in `tests/testsuite/edit.rs` covering the `[table]` case (plus a root-only case, a nested/array-of-tables case, and empty input), asserting `Document<String>::to_string()` equals the source and matches `DocumentMut`. Full `cargo test -p toml_edit -p toml` passes, `cargo fmt --check` is clean, and there are no new clippy warnings.

I saw your note that this needs more thought on how it should be fixed, so I'm happy to adjust the approach if you'd prefer different semantics (e.g. a different `Display` for the immutable document, or not exposing `to_string()` on it at all).

Fixes #1008.

---

Disclosure: this fix was developed with AI assistance; I reviewed it and stand behind it.

### DIFF
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -6,6 +6,7 @@ use toml_writer::ToTomlValue as _;
 use toml_writer::TomlWrite as _;
 
 use crate::DocumentMut;
+use crate::RawString;
 use crate::inline_table::DEFAULT_INLINE_KEY_DECOR;
 use crate::key::Key;
 use crate::repr::{Decor, Formatted, Repr, ValueRepr};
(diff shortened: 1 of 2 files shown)


epage:
> Disclosure: this fix was developed with AI assistance; I reviewed it and stand behind it.

The challenge when too much AI boilerplate is used, it injects doubt in how much I'll be interacting with a human vs an AI, particularly when "obviously" incorrect things are present (though I recognize that my bias as a maintainer is that I am more familiar with what "obvious" means).

epage:
Also, our contrib guide asks for getting buy-in on a solution before posting PRs.

hdimer:
Fair on both counts. You mentioned there was a reason `Display` was left off. Rather than guess, how would you like `to_string()` on `Document<S>` to behave? Happy to go whichever way you'd prefer.

epage:
Thats a question for the issue, per our contrib guide, and is why it is still open. Most of what is needed in open source is thinking, not code.

hdimer:
Makes sense, I got ahead of the process here. Closing this; happy to pick up the implementation once the design is settled on #1008.

PR #1197, opened 2026-07-30
feat: Add `ValueSerializer::pretty`

rossmacarthur:
Previously there wasn't a way to serialize toml *values* with a "pretty" style, only documents. This adds a `ValueSerializer::pretty` method that creates a serializer with the default "pretty" style.


epage:
Note that our contribution guide asks for Issue for non-trivial changes before posting a PR. These issues should be focused on your use case / the problem you are trying to solve.

Moving to a Draft until then.

PR #1215, opened 2026-09-08
perf: Reduce allocations in toml_edit parsing and dumping

dimbleby:
Some rather modest changes aimed at performance.  parse time doesnt really change, but I see dump time improve by 30-40%

---

Borrow input and move completed header keys when parsing documents. Borrow header keys when dumping, move generated representations, and write existing strings directly to avoid copies and formatting overhead.

Add dump and edit-roundtrip benchmarks with allocation profiling, plus regression coverage for input ownership and write errors.

### DIFF
--- a/crates/benchmarks/Cargo.toml
+++ b/crates/benchmarks/Cargo.toml
@@ -41,5 +41,9 @@ harness = false
 name = "2-array"
 harness = false
 
+[[bench]]
+name = "3-keys"
+harness = false
+
 [lints]
 workspace = true
--- a/crates/benchmarks/Cargo.toml
+++ b/crates/benchmarks/Cargo.toml
@@ -10,6 +10,7 @@ release = false
 
 [features]
 default = []
+alloc-profiler = []
 simd = ["toml_parser/simd"]
 unsafe = ["toml_parser/unsafe"]
 preserve_order = ["serde_json/preserve_order", "toml_old/preserve_order", "toml/preserve_order"]
--- a/crates/benchmarks/benches/0-cargo.rs
+++ b/crates/benchmarks/benches/0-cargo.rs
@@ -1,5 +1,9 @@
 #![allow(elided_lifetimes_in_paths)]
 
+#[cfg(feature = "alloc-profiler")]
+#[global_allocator]
+static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();
+
 mod toml_parser {
     use toml_benchmarks::{Data, MANIFESTS};
 
--- a/crates/benchmarks/benches/0-cargo.rs
+++ b/crates/benchmarks/benches/0-cargo.rs
@@ -93,6 +93,15 @@ mod toml_edit {
     }
 
     #[divan::bench(args=MANIFESTS)]
+    fn dump(bencher: divan::Bencher, sample: &Data<'static>) {
+        let document = sample
+            .content()
+            .parse::<::toml_edit::DocumentMut>()
+            .unwrap();
+        bencher.bench(|| std::hint::black_box(&document).to_string());
+    }
+
+    #[divan::bench(args=MANIFESTS)]
     fn manifest(sample: &Data<'static>) -> manifest::Manifest {
         ::toml_edit::de::from_str(sample.content()).unwrap()
     }
@@ -122,6 +131,12 @@ mod toml {
     }
 
     #[divan::bench(args=MANIFESTS)]
+    fn dump(bencher: divan::Bencher, sample: &Data<'static>) {
+        let document = sample.content().parse::<::toml::Table>().unwrap();
+        bencher.bench(|| ::toml::to_string(std::hint::black_box(&document)).unwrap());
+    }
+
+    #[divan::bench(args=MANIFESTS)]
     fn manifest(sample: &Data<'static>) -> manifest::Manifest {
         ::toml::de::from_str(sample.content()).unwrap()
     }
@@ -133,6 +148,12 @@ mod toml_v05 {
     #[divan::bench(args=MANIFESTS)]
     fn document(sample: &Data<'static>) -> ::toml_old::Value {
         sample.content().parse().unwrap()
+    }
+
+    #[divan::bench(args=MANIFESTS)]
+    fn dump(bencher: divan::Bencher, sample: &Data<'static>) {
+        let document = sample.content().parse::<::toml_old::Value>().unwrap();
+        bencher.bench(|| ::toml_old::to_string(std::hint::black_box(&document)).unwrap());
     }
 
     #[divan::bench(args=MANIFESTS)]
@@ -156,6 +177,12 @@ mod serde_json {
             .bench_values(|sample| {
                 serde_json::from_str::<::toml::Value>(sample.content()).unwrap()
             });
+    }
+
+    #[divan::bench(args=MANIFESTS)]
+    fn dump(bencher: divan::Bencher, sample: &Data<'static>) {
+        let document = toml_edit::de::from_str::<::serde_json::Value>(sample.content()).unwrap();
+        bencher.bench(|| ::serde_json::to_string(std::hint::black_box(&document)).unwrap());
     }
 
     #[divan::bench(args=MANIFESTS)]
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -33,7 +33,7 @@ pub(crate) fn encode_key(this: &Key, buf: &mut dyn Write, input: Option<&str>) -
 }
 
 fn encode_key_path(
-    this: &[Key],
+    this: &[&Key],
     mut buf: &mut dyn Write,
     input: Option<&str>,
     default_decor: (&str, &str),
@@ -214,7 +214,7 @@ impl Display for DocumentMut {
             if let Some(pos) = t.position() {
                 last_position = pos;
             }
-            tables.push((last_position, t, p.clone(), is_array));
+            tables.push((last_position, t, p.to_vec(), is_array));
             Ok(())
         })
         .unwrap();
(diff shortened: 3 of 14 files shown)


epage:
Please see our contrib guide for PR requirements.

dimbleby:
> Please see our contrib guide for PR requirements.

Looks fine to me.  Please spell out for me what change you are asking for

dimbleby:
my best guess is that you want a longer commit history?  though there is a large gap between "ideals" in the contributing guide and "requirements" here, so maybe I have guessed wrong and still missed something.

epage:
> my best guess is that you want a longer commit history? though there is a large gap between "ideals" in the contributing guide and "requirements" here, so maybe I have guessed wrong and still missed something.

Not longer commit history but atomic commit history. As an author, you have the context of how this is all put together but then that gets thrown out by putting everything into a single commit, on top of other reasons like bisecting.

An "ideal" is something to strive for but there may be reasons that you can't fully live up to it. That means it isn't something to ignore.

crates/benchmarks/benches/0-cargo.rs
```diff
@@ -1,5 +1,9 @@
 #![allow(elided_lifetimes_in_paths)]
 
+#[cfg(feature = "alloc")]
+#[global_allocator]
+static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();
+
```
epage: >  Allow allocation profiling through the benchmark crate's
alloc feature so the later optimizations have a reproducible baseline.

This feels distinct from roundtrip benchmarking and could be its own commit.

crates/benchmarks/benches/0-cargo.rs
```diff
@@ -88,6 +92,28 @@ mod toml_edit {
...
     }
 
+    #[divan::bench(args=MANIFESTS)]
+    fn dump(bencher: divan::Bencher, sample: &Data<'static>) {
+        let document = sample
+            .content()
+            .parse::<::toml_edit::DocumentMut>()
+            .unwrap();
+        bencher.bench(|| std::hint::black_box(&document).to_string());
+    }
```
epage, line 102: This can be generalized to work in other mods within these files

crates/benchmarks/benches/0-cargo.rs
```diff
@@ -88,6 +92,28 @@ mod toml_edit {
...
+        let mut document = sample
+            .content()
+            .parse::<::toml_edit::DocumentMut>()
+            .unwrap();
+        let version = document["package"]["version"].as_value_mut().unwrap();
+        let decor = version.decor().clone();
+        *version = ::toml_edit::Value::from("0.2.0");
+        *version.decor_mut() = decor;
+        document.to_string()
+    }
```
epage: What is it we are expecting to better understand with this benchmarked compared to parse and render ones?

crates/benchmarks/Cargo.toml
```diff
@@ -10,6 +10,7 @@ release = false
 
 [features]
 default = []
+alloc = []
```
epage: I feel like this needs a comment or a better name because `alloc` generally refers to the [alloc](https://doc.rust-lang.org/stable/alloc/) standard library

crates/toml_edit/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 <!-- next-header -->
 ## [Unreleased] - ReleaseDate
 
+### Performance
+
+- Avoid copying the input when parsing editable documents
+
```
epage: I'd prefer changelogs to not be updated as part of PRs. It just creates a source of conflicts. It is fairly easy to write up a changelog from good commits

crates/toml_edit/src/parser/document.rs
```diff
@@ -346,7 +346,7 @@ impl State {
         let _scope = TraceScope::new("document::finish_table");
         let mut prev_table = std::mem::take(&mut self.current_table);
         if let Some(header) = self.current_header.take() {
-            let Some(key) = &header.key else {
+            let Some(key) = header.key else {
```
epage, line 349: I'll need to revisit this to see how this is related to the `toml` parser.

The two parsers are intentionally kept to be diffable. I depending on if can be carried over or how the diff looks, I would be willing to sacrifice performance here. Performance for edits is not usually in a critical path like it is for non-edit parses.

crates/toml_edit/src/encode.rs
```diff
@@ -65,33 +65,12 @@ fn encode_key_path(
...
-        encode_key(key, buf, input)?;
-
-        if last {
-            leaf_decor.suffix_encode(buf, input, default_decor.1)?;
-        } else {
-            dotted_decor.suffix_encode(buf, input, DEFAULT_KEY_PATH_DECOR.1)?;
-        }
-    }
-    Ok(())
+    encode_key_path(this, buf, input, default_decor, leaf_decor)
```
epage: Instead of making this change in this commit, consider having a follow up commit that is a refactor to remove `encode_key_path_ref`

epage:
Can you measure the impact of each commit? For example, I'm wondering if ce0297cbc0f69029a07f451e636b4488939486ab is really worth it.

dimbleby:
> Can you measure the impact of each commit? For example, I'm wondering if [ce0297c](https://github.com/toml-rs/toml/commit/ce0297cbc0f69029a07f451e636b4488939486ab) is really worth it.
```
┌──────────────────────────┬──────────────────────────────────────────────────────────────┬──────────────────────────────┐
│ Commit                   │ Timing result                                                │ Allocation result            │
├──────────────────────────┼──────────────────────────────────────────────────────────────┼──────────────────────────────┤
│ a44dec8a — borrow input  │ No convincing general parsing speedup                        │ Removes one whole-input      │
│                          │                                                              │ allocation                   │
├──────────────────────────┼──────────────────────────────────────────────────────────────┼──────────────────────────────┤
│ 6b44c4da — move parser   │ Small, mixed parsing effects                                 │ Removes 100 allocations when │
│ header keys              │                                                              │ parsing 100 tables           │
├──────────────────────────┼──────────────────────────────────────────────────────────────┼──────────────────────────────┤
│ 02087ac5 — borrow        │ About 17% / 9% / 2% on small/medium/large manifests; 25–30%  │ Medium dump: 238 → 182;      │
│ dump-header keys         │ on table-heavy fixtures                                      │ 100-table dump: 805 → 405    │
├──────────────────────────┼──────────────────────────────────────────────────────────────┼──────────────────────────────┤
│ aeee9b52 — move          │ About 7% when dumping freshly constructed keys/values;       │ 100-entry generated          │
│ generated                │ unchanged-document dumps effectively unchanged               │ document: 504 → 304          │
│ representations          │                                                              │                              │
├──────────────────────────┼──────────────────────────────────────────────────────────────┼──────────────────────────────┤
│ ce0297cb — direct string │ 24–30% on parsed-manifest dumps                              │ No allocation reduction      │
│ writes                   │                                                              │                              │
└──────────────────────────┴──────────────────────────────────────────────────────────────┴──────────────────────────────┘
```
Direct string writes is the largest single win

crates/benchmarks/benches/0-cargo.rs
dimbleby: not a lot probably, removed

crates/benchmarks/Cargo.toml
dimbleby: Please prefer to be explicit about what change you would like, guessing is stressful and more likely to produce results that you still do not like.

I have rolled the dice on `alloc-profiler`

crates/toml_edit/src/parser/document.rs
dimbleby: made the same change in both places

crates/toml_edit/src/encode.rs
dimbleby: Please prefer to be direct about what you want to happen.  I have assumed that you did not only want me to "consider" this, and made the change..

crates/benchmarks/benches/0-cargo.rs
dimbleby: "could".  

I have assumed that you meant "please make this its own commit" and done that

crates/benchmarks/benches/0-cargo.rs
dimbleby: "can" - again I have assumed that you intended that I do this: and done it.

crates/toml_edit/CHANGELOG.md
dimbleby: removed

crates/benchmarks/Cargo.toml
epage: > Please prefer to be explicit about what change you would like, guessing is stressful and more likely to produce results that you still do not like.

1. Reviews are a two way conversation. I'm not speaking in absolutes and dictating exactly what should be done because I want to leave room to hear from you if there is a reason or it otherwise
2. If I were to take the time an think through all of the details and give very precise directions, that isn't much different than me having an agent make the change or for me to make it myself. It is a big help for contributors to think through these things and lead out in what should be done.

crates/toml_edit/src/parser/document.rs
epage: Thank!

crates/benchmarks/benches/0-cargo.rs
```diff
@@ -101,6 +101,26 @@ mod toml_edit {
         bencher.bench(|| std::hint::black_box(&document).to_string());
     }
 
+    #[divan::bench(args = [10, 100])]
+    fn dump_generated(bencher: divan::Bencher, entries: usize) {
+        let mut document = ::toml_edit::DocumentMut::new();
```
epage: This was put in `0-cargo` which doesn't really fit since this is unrelated to those real-world benchmarks.

Maybe we should add a synthetic benchmark file for keys, like `3-keys.rs`

epage:
Thanks for the clean ups, this is looking great! Oh well about dropping the direct write change; that benchmark is too meaningful to ignore.

crates/benchmarks/benches/0-cargo.rs
dimbleby: I am neutral on this, if you leave it at "maybe" then I wont bother but I am quite happy to shuffle it around if you think it worthwhile.

epage: The "maybe" was on one idea on how to resolve this.

dimbleby: Done.  Apparently we are only barely speaking the same language!

epage:
Thanks!

PR #1216, opened 2026-09-10
perf: reduce allocations dumping array-heavy documents

dimbleby:
Adds concrete internal iterators to `Array` and `ArrayOfTables`

Greatly reduces allocations in array-heavy documents (eg feature-heavy Cargo.toml), perf win maybe 10% on favourable documents


crates/toml_edit/src/array.rs
```diff
@@ -106,7 +106,11 @@ impl Array {
 impl Array {
     /// Returns an iterator over all values.
     pub fn iter(&self) -> ArrayIter<'_> {
-        Box::new(self.values.iter().filter_map(Item::as_value))
+        Box::new(self.iter_values())
+    }
+
+    pub(crate) fn iter_values(&self) -> impl Iterator<Item = &Value> {
+        self.values.iter().filter_map(Item::as_value)
     }
```
epage, line 114: I wonder if the API dates back to before return position impl trait and we've just not remembered to switch to it when making a breaking change.

epage:
Thanks!

PR #1217, opened 2026-09-10
perf: remove more allocations in dumping

dimbleby:
This one is more intrusive than the just-merged PR, it is not so clear to me that the juice is worth the squeeze.

Idea is to replace recursive collection in `Table.get_values()` (and `InlineTable.get_values()`) with a `ValueEntries` cursor.

Commits set up:
- behaviour tests that the refactor should preserve
- benchmarks for documents that will show improvement
- preparatory refactor introducing the cursor
- use the cursor, for maybe a 15-20% win in typical documents

### DIFF
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -32,18 +32,38 @@ pub(crate) fn encode_key(this: &Key, buf: &mut dyn Write, input: Option<&str>) -
     Ok(())
 }
 
-pub(crate) fn encode_key_path(
+fn encode_key_path(
     this: &[&Key],
+    buf: &mut dyn Write,
+    input: Option<&str>,
+    default_decor: (&str, &str),
+    leaf_decor: &Decor,
+) -> Result {
+    let (leaf, prefix) = this.split_last().expect("always at least one key");
+    encode_key_path_parts(
+        prefix.iter().copied(),
+        leaf,
+        buf,
+        input,
+        default_decor,
+        leaf_decor,
+    )
+}
+
+fn encode_key_path_parts<'k>(
+    prefix: impl ExactSizeIterator<Item = &'k Key>,
+    leaf: &'k Key,
     mut buf: &mut dyn Write,
     input: Option<&str>,
     default_decor: (&str, &str),
     leaf_decor: &Decor,
 ) -> Result {
-    for (i, key) in this.iter().enumerate() {
+    let leaf_index = prefix.len();
+    for (i, key) in prefix.chain(std::iter::once(leaf)).enumerate() {
         let dotted_decor = key.dotted_decor();
 
         let first = i == 0;
-        let last = i + 1 == this.len();
+        let last = i == leaf_index;
 
         if first {
             leaf_decor.prefix_encode(buf, input, default_decor.0)?;
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -324,7 +324,18 @@ fn visit_table(
         table.decor.suffix_encode(buf, input, default_decor.1)?;
         writeln!(buf)?;
     }
-    // print table body
+    encode_table_values(children, buf, input)
+}
+
+pub(crate) fn encode_table_body(table: &Table, buf: &mut dyn Write, input: Option<&str>) -> Result {
+    encode_table_values(table.get_values(), buf, input)
+}
+
+fn encode_table_values(
+    children: Vec<(Vec<&Key>, &Value)>,
+    mut buf: &mut dyn Write,
+    input: Option<&str>,
+) -> Result {
     for (key_path, value) in children {
         let leaf_decor = key_path
             .last()
--- a/crates/toml_edit/src/encode.rs
+++ b/crates/toml_edit/src/encode.rs
@@ -11,6 +11,7 @@ use crate::key::Key;
 use crate::repr::{Decor, Formatted, Repr, ValueRepr};
 use crate::table::{
     DEFAULT_KEY_DECOR, DEFAULT_KEY_PATH_DECOR, DEFAULT_ROOT_DECOR, DEFAULT_TABLE_DECOR,
+    ValueEntries,
 };
 use crate::value::{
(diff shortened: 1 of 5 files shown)


crates/toml_edit/tests/testsuite/edit.rs
```diff
@@ -130,6 +132,75 @@ fn document_display_propagates_write_errors() {
     assert_eq!(writer.0, 0);
 }
 
+#[test]
+fn visual_values_skip_empty_and_standard_tables() {
+    let input = "first.a = 1\nsecond.b = 2\nplain = 3\n";
```
epage, line 137: Please avoid escaped values, preferring to put content in a format that reads more a naturally.

crates/toml_edit/tests/testsuite/edit.rs
```diff
@@ -130,6 +132,75 @@ fn document_display_propagates_write_errors() {
...
+    table.insert("empty", Item::Table(empty));
+    table.insert("none", Item::None);
+    let mut nested = Table::new();
+    nested.insert("not_in_body", value(4));
+    table.insert("header_only", Item::Table(nested));
+
+    assert_eq!(
+        visual_paths(&table),
+        vec![vec!["first", "a"], vec!["second", "b"], vec!["plain"]]
+    );
```
epage, line 150: What is the goal with this vs the assert on the rendered content?

crates/toml_edit/src/table.rs
```diff
@@ -823,3 +857,212 @@ impl<'a> VacantEntry<'a> {
...
+                    reference_inline(table, path, values);
+                }
+                Item::Value(value) => values.push((path.clone(), value)),
+                _ => {}
+            }
+            path.pop();
+        }
+    }
+
+    proptest! {
```
epage, line 1020: This is a pretty complicated prop test and I worry about the long term maintainablity of it.

Is it worth it?

crates/toml_edit/src/encode.rs
```diff
@@ -32,18 +33,20 @@ pub(crate) fn encode_key(this: &Key, buf: &mut dyn Write, input: Option<&str>) -
     Ok(())
 }
 
-pub(crate) fn encode_key_path(
-    this: &[&Key],
+fn encode_key_path<'k>(
```
epage, line 36: `encode_key_path_ref` took over the existing signature for this function but the names don't distinguish there roles. The `_ref` suffix before was to clarify owned vs borrowed but here its keys vs prefix+leaf

crates/toml_edit/src/table.rs
```diff
@@ -422,19 +421,7 @@ impl Table {
...
-                .last()
-                .expect("always at least one key")
-                .leaf_decor();
-            crate::encode::encode_key_path(&key_path, f, None, DEFAULT_KEY_DECOR, leaf_decor)?;
-            write!(f, "=")?;
-            crate::encode::encode_value(value, f, None, DEFAULT_VALUE_DECOR)?;
-            writeln!(f)?;
-        }
-        Ok(())
+        crate::encode::encode_table_body(self, f, None)
```
epage, line 424: If a move like this is being made, it should be done in a separate refactor commit so the changes in the commit can focus purely on the titled change, see also https://epage.github.io/dev/pr-style/#c-split

crates/toml_edit/src/encode.rs
```diff
@@ -319,18 +348,25 @@ fn visit_table(
...
     }
-    // print table body
-    for (key_path, value) in children {
-        let leaf_decor = key_path
-            .last()
-            .expect("always at least one key")
-            .leaf_decor();
-        encode_key_path(&key_path, buf, input, DEFAULT_KEY_DECOR, leaf_decor)?;
+    encode_table_values(&mut children, buf, input)
+}
```
epage, line 357: Does the splitting out of `encode_table_values` need to be on this commit?

See also https://epage.github.io/dev/pr-style/#c-split

epage: Note for these commit structure items: to be clear, this isn't busy work. I've stopped reviewing at this point because I don't wan to take the time to reverse engineer the changes you made to make sure I fully understand and agree with them when you have that state already and, with some changes in commit structure, can share that state with me.

crates/toml_edit/src/table.rs
dimbleby: Well, I'm in two minds about whether the whole thing is worth it.  The cursor is quite complicated and plausibly a maintenance challenge.

This test is intended to convince that the cursor implementation matches the simpler code.

If you are sufficiently convinced of that: then perhaps it already has served its purpose.  I'd have no objection to removing it if you prefer.

crates/toml_edit/src/encode.rs
dimbleby: meh, the bot has plenty of patience for this sort of thing, if it - and I - can only be taught to understand what you are after.

crates/toml_edit/src/table.rs
epage: Overall, the Cursor doesn't seem too complicated. The worst parts are
- Some of the conditions are a bit more decoupled from the state they are derived from (`flatten_all_tables`)
- ... I already forgot what the other part was which just shows how important it was

As for future changes this may interfere with, #163 is the main one I can think of since we'd need to do sorting on top of what gets returned. For the use cases I care about (`cargo add`), #163 isn't a big deal so I'm in no rush to fix it.

crates/toml_edit/tests/testsuite/edit.rs
dimbleby: the idea is to assert something about `.get_values()`, but clearer perhaps to do that more separately.

crates/toml_edit/src/encode.rs
dimbleby: agree, `_ref` is unhelpful naming - changed it

crates/toml_edit/src/encode.rs
dimbleby: inserted a couple more intermediate commits.

crates/toml_edit/src/table.rs
dimbleby: I - or my AI friend - am quite happy either to switch in a handful of deterministic tests on cursor behaviour, or to remove the proptest altogether.  Let us know what suits you.

PR #1219, opened 2026-09-13
fix(edit): normalize CRLF newlines in multi-line strings

### LINKED ISSUE #1213: A CRLF document with a multi-line string prints back with mixed line endings
#1158 raised the possibility that the parser rewrite preserves CRLF where it explicitly tracks
newlines, and so "may be causing mixed newline characters to be used", and said discussion of CRLF
should be its own issue. This is that issue: an unedited document whose line endings are all CRLF
prints back with CRLF inside the multi-line string and LF everywhere else.

```rust
use toml_edit::DocumentMut;

fn main() {
    let with_multiline = "s = \"\"\"\r\ntext\r\n\"\"\"\r\nb = 2\r\n";
    let without = "a = 1\r\nb = 2\r\n";

    for src in [with_multiline, without] {
        println!("in:  {src:?}");
        println!("out: {:?}", src.parse::<DocumentMut>().unwrap().to_string());
    }
}
```

Output:

```
in:  "s = \"\"\"\r\ntext\r\n\"\"\"\r\nb = 2\r\n"
out: "s = \"\"\"\r\ntext\r\n\"\"\"\nb = 2\n"
in:  "a = 1\r\nb = 2\r\n"
out: "a = 1\nb = 2\n"
```

The second document has no multi-line string and comes back uniformly LF, so a construct whose
newlines are kept verbatim is what it takes to get both endings into one file. The `Limitations`
list added when #1158 was closed covers the lack of a trailing newline and the order of dotted keys,
not this.

I expected an unedited document to print back byte for byte, or failing that for the output's line
endings to be uniform.

Tested on toml_edit 0.25.13+spec-1.1.0 and on `main` at 1189a12.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

devtechedge:
An all-CRLF document with a multi-line string roundtrips with mixed line endings: CRLF inside the string and LF everywhere else. Structural newlines are re-emitted as LF, so a construct whose newlines are kept verbatim is all it takes to get both endings into one output file.

Structural newlines are re-emitted from the parser's own newline tracking, while multi-line string content was decoded verbatim, CRLF included.

The TOML 1.0.0 spec permits normalization here: the multi-line basic strings section says "TOML parsers should feel free to normalize newline to whatever makes sense for their platform", and multi-line literal strings "must normalize newlines in the same manner as multi-line basic strings". This normalizes CRLF to LF when decoding multi-line basic and multi-line literal strings, and re-encodes a string's raw representation with the same normalization so it stays consistent with the decoded value. An unedited document now roundtrips with uniform line endings.

Tests:

- `crlf_multiline_string_in_crlf_document_roundtrips` and `crlf_multiline_literal_string_in_crlf_document_roundtrips` fail before and pass after
- `crlf_document_without_multiline_string_roundtrips_uniformly` and `lf_multiline_string_in_crlf_document_roundtrips_uniformly` pin the already-uniform cases
- decoder unit tests cover CRLF content in `toml_parser`

Fixes #1213


devtechedge:
Hi there - gentle nudge on the CRLF multi-line string roundtrip whenever you have a moment.

An all-CRLF document with a multi-line string was roundtripping with mixed endings: CRLF kept inside the string while structural newlines re-emitted as LF (`Fixes #1213`).

This normalizes CRLF to LF when decoding multi-line basic and multi-line literal strings in `crates/toml_parser/src/decoder/string.rs`, and re-encodes the string raw repr with the same normalization in `crates/toml_edit/src/parser/value.rs`, which the TOML 1.0.0 multi-line string rules permit.

Tests in `crates/toml_edit/tests/testsuite/edit.rs` pin the failing roundtrip cases (`crlf_multiline_string_in_crlf_document_roundtrips` and the literal variant) plus already-uniform controls; decoder unit coverage sits alongside in `toml_parser`.

The branch follows the repo CONTRIBUTING shape with a test commit first, then the fix commit; local `cargo test --workspace`, fmt, clippy, and docs checks were green (fuzz crate excluded locally for clang).

Glad to split the decoder vs edit changes differently, or tighten the normalization scope, if that would ease review.

Whenever it fits is fine - still ready from my end.

